PR #69, opened 2026-01-14
Handle error state for generating transactions correctly

0xnullifier:


WiktorStarczewski:
Closing as superseded by #193, which rewrote GeneratingTransaction.tsx with a per-stage progress model. The error-gating bug this PR fixed (error icon showing before transactionComplete) no longer applies — the component now uses hasErrors + failedCount driven by getFailedTransactions() polling and correctly gates the failure icon on transactionComplete && hasErrors. The MAX_WAIT_BEFORE_CANCEL change is also no longer relevant: main now has a mobile/desktop split (2min / 30min) in seconds rather than ms.

Issue #64, opened 2026-01-12
Allow for exporting/importing of accounts through Private Keys and Multiple Seed phrases in a single wallet


0xnullifier:
Currently there is no way for the users to export and import accounts for wallet with a private key.

It is not as simple and would require some changes to how we do things with the seed AFAIK we should change and use the seed from bip39 for only secret key generation.

We currently use `client.newWallet` for creating a new account but this uses seed for both secret key generation and account seed thus exporting an account would require either the full seed or secret key + account id to import the account.

As for Multiple seed phrase it is open for discussion, some wallets do it some don't.

PR #76, opened 2026-01-14
Enable private key export and import

### LINKED ISSUE #115: Miden Wallet not able to retrieve private key
### Feature description

Add an option that allow us to retrieve private key/secret key for specific account instead of only seed phrase

### Why is this feature needed?

Currently there's no way for us to retrieve our private key for specific account and only seedphrase, in case we only need to import specific account its not doable right now

0xnullifier:
Closes one part of #64 

It changes the seed phrase derived client seed to be used for only secret key generation and thus enables export and import purely based on private keys

Closes [Feedback Issue](https://github.com/0xMiden/feedback/issues/115
)


WiktorStarczewski:
Superseded by #195, which carries just the unique private-key export/import work from this branch reimplemented against main's current SDK shape (the React SDK v14 migration in #151 and the webpack→Vite + E2E harness in #182 landed after this branch diverged, and about half of this PR's commits shipped separately via #78, so a clean rebase wasn't viable). #195 addresses the reviewer feedback on top of that, including the encrypted-wallet-file corruption that this PR's `hdIndex: -1` accounts would have triggered.

PR #77, opened 2026-01-15
Fix Auto Consume behaviour in case of failed transaction

0xnullifier:


WiktorStarczewski:
Closing as obsolete. Main's PR #184 added dedup in initiateConsumeTransaction that addresses the duplicate consume issue, but with the opposite failure policy (Failed txs are explicitly excluded so retries work — see transactions.ts:158-159). This branch's 'skip on prior failure' approach conflicts with that design and is also 3 months stale + conflicting.

PR #128, opened 2026-02-16
Update README.md

S0nee:
1. Corrected incorrect paths:
- `createWalletFlow/Welcome.tsx` -> `common/Welcome.tsx`
- `createWalletFlow` -> `create-wallet-flow`
- `importWalletFlow` -> `import-wallet-flow`

2. For greater convenience, I immediately added links so that readers can quickly navigate to the files they need.

### DIFF
--- a/src/screens/onboarding/README.md
+++ b/src/screens/onboarding/README.md
@@ -14,35 +14,35 @@ The onboarding feature guides the user through setting up their wallet in the ap
 
 ### Welcome Screen
 
-- **Path**: `createWalletFlow/Welcome.tsx`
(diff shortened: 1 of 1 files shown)


WiktorStarczewski (changes requested):
Thanks for cleaning these up — the renamed paths all check out against the current tree.

One change needed before merge: the links point to your fork (`github.com/S0nee/miden-wallet/blob/README/...`) rather than upstream. Once this branch/fork goes away those links 404. For an in-repo README, relative paths are the standard and survive renames automatically:

- [`common/Welcome.tsx`](./common/Welcome.tsx)
- [`create-wallet-flow/BackUpSeedPhrase.tsx`](./create-wallet-flow/BackUpSeedPhrase.tsx)

...etc. Could you swap them over?

PR #133, opened 2026-02-21
fix: note consume issues: batch consume, mutex contention, stuck transactions

zacnider:
## Summary

This PR fixes three critical note consume issues in the Miden Wallet Chrome extension:

1. **Notes from faucet fail to consume or consume very late** - AutoSync (3s polling) and claimable notes polling (5s) compete with consume transactions for the single WASM AsyncMutex, starving consume operations
2. **Multiple incoming notes pile up in the queue** - Each note creates a separate ConsumeTransaction, processing sequentially through the mutex bottleneck  
3. **Stuck transactions block the entire queue** - A transaction stuck in `GeneratingTransaction` state blocks all subsequent transactions indefinitely (30min timeout on desktop)

### Fix 1: Pause AutoSync during consume on all platforms
- Removed the `isMobile()` guard so AutoSync pauses on desktop too when transaction modal is open

### Fix 2: Pause claimable notes polling during consume
- Set SWR `refreshInterval` to 0 when `isTransactionModalOpen` is true
- Updated `notesBeingClaimed` filter to support batch `noteIds`

### Fix 3: Batch consume - multiple notes in a single transaction
- `ConsumeTransaction` now accepts `ConsumableNote[]` and stores `noteIds: string[]`
- SDK interface uses full `noteIds` array (SDK already supports `Note[]`)
- Added `initiateBatchConsumeTransaction()` for single batch transaction
- "Claim All" creates one transaction instead of N separate ones

### Fix 4: Auto-cancel stuck transactions in processing loop
- Detects transactions stuck in `GeneratingTransaction` state and auto-cancels them

## Test plan

- [x] Request tokens from faucet → verify note is consumed quickly
- [x] Multiple notes → "Claim All" creates a single batch transaction
- [x] Verify AutoSync pauses during consume (console log)
- [x] Stuck transaction auto-cancels after ~1 minute
- [x] Build succeeds: `yarn build:extension`

### DIFF
--- a/src/app/pages/Receive.tsx
+++ b/src/app/pages/Receive.tsx
@@ -15,6 +15,7 @@ import { formatBigInt } from 'lib/i18n/numbers';
 import {
   getFailedTransactions,
   getUncompletedTransactions,
+  initiateBatchConsumeTransaction,
   initiateConsumeTransaction,
   verifyStuckTransactionsFromNode,
   waitForConsumeTx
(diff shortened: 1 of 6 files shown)


WiktorStarczewski:
Closing — the issues this targets have been addressed on main by #171 (consume dedup) and #150 (graceful transaction recovery). The branch is also significantly behind main (22 commits, including the React SDK + Miden Client v14 migration in #151 and the webpack→Vite migration in #182), and one of the files it modifies (src/lib/miden/front/autoSync.ts) no longer exists. Thanks for the contribution!

PR #150, opened 2026-03-05
fix: graceful transaction recovery after network outages

### LINKED ISSUE #149: Private accounts can enter unrecoverable state after network outage
### What should be done?

When a transaction is submitted for a private account and the node commits it but the local client doesn't update its state (e.g., due to network outage, extension crash, or service worker termination), the account becomes permanently stuck. All subsequent transactions fail with: `transaction's initial state commitment does not match the account's current value`

`syncState()` cannot fix this because the node only stores the state commitment hash for private accounts, not the full account state. The local state is the only copy and it's now stale.

Additionally, several contributing issues make recovery harder:

1. **No sync-before-execute:** `generateTransaction` builds and submits transactions without ensuring the local state is up-to-date first. AutoSync runs on a separate 3s loop with no coordination.
2. **AutoSync stops permanently on generating-transaction page:** The `sync()` method `return`s without re-scheduling itself when `generating-transaction` is in the URL (line 72-76 of `autoSync.ts`), and never restarts until the next page navigation triggers `updateState()`.
3. **Stale queued transactions have no TTL:** Queued transactions (`status === 0`) survive indefinitely in IndexedDB across sessions and outages. There's no expiry based on `initiatedAt`. `cancelStuckTransactions` only cleans up in-progress transactions.
4. **In-progress transactions with undefined `processingStartedAt` are never cancelled:** `cancelStuckTransactions` filters on `tx.processingStartedAt && Date.now() - tx.processingStartedAt > MAX_WAIT_BEFORE_CANCEL` — if `processingStartedAt` is undefined (crash before it was set), the expression is falsy and the tx is skipped forever.
5. **`displayMessage` not updated on failure:** `cancelTransaction` sets `status = Failed` but never updates `displayMessage`, so failed transactions still show "Executing" / "Sending" in the UI.

### How should it be done?

1. **Sync before executing transactions:** In `generateTransaction`, call `syncState()` before building the transaction to ensure local state matches the node. This is the most critical fix.
2. **Fix AutoSync stopping permanently:** Replace the bare `return` on line 75 of `autoSync.ts` with `await sleep(3000); await this.sync();` so the sync loop continues even when on the generating-transaction page (just skip the actual sync call).
3. **Add TTL for queued transactions:** In `cancelStuckTransactions` (or a new function), also cancel queued transactions older than a threshold (e.g., 30 minutes). Stale queued txs will almost certainly fail anyway.
4. **Handle undefined `processingStartedAt`:** Change the filter in `cancelStuckTransactions` to also cancel in-progress txs where `processingStartedAt` is undefined — these are transactions that crashed before processing started and will never complete.
5. **Update `displayMessage` on cancellation:** In `cancelTransaction`, set `dbTx.displayMessage` to a failure message (e.g., "Failed") so the history UI reflects the actual status.
6. **For private account recovery (longer-term):** Persist the `TransactionResult` locally before submitting to the node, so the new state can be derived even if the submission response is lost. This would allow replaying the state transition on the next sync.

### When is this task done?

-

### Additional context

_No response_

WiktorStarczewski:
Closes #149

## Summary

- **Fix AutoSync loop death**: The sync loop was permanently killed when visiting the generating-transaction page. Now it skips the sync call (avoiding lock contention) but stays alive and resumes once navigating away.
- **Sync before execute**: Calls `syncState()` before generating transactions to ensure the local WASM client has the latest account state, preventing "initial state commitment does not match" errors after network outages.
- **Cancel crashed transactions**: Transactions stuck in `GeneratingTransaction` with undefined `processingStartedAt` (app crash during transition) are now properly cancelled.
- **Better cancellation UX**: Cancelled transactions now show "Failed" status with proper icon instead of misleading "Executing".
- **Queued transaction TTL**: Transactions queued for over 30 minutes are automatically cancelled, preventing stale tx buildup.
- **Modal resilience**: The transaction progress modal no longer blocks on a single stale transaction failure — it continues processing remaining queued transactions.

## Test plan

- [x] All 891 existing tests pass
- [x] 8 new tests added covering: loop survival, sync-before-execute ordering, undefined processingStartedAt handling, error serialization, queued TTL, and stale tx preservation
- [x] Lint and format clean
- [ ] Manual: build extension, send transaction, verify syncState called before generation in service worker console

### DIFF
--- a/src/lib/miden/activity/transactions.test.ts
+++ b/src/lib/miden/activity/transactions.test.ts
@@ -558,6 +558,202 @@ describe('transactions utilities', () => {
   });
 });
 
+/**
+ * Integration test: full network outage recovery flow.
+ * Uses jest.isolateModules with a stateful in-memory DB to simulate:
+ *   1. Network up → transaction succeeds
+ *   2. Network down → syncState fails → transaction cancelled
+ *   3. Network back up → new transaction succeeds
+ */
+describe('Transaction resilience: network outage recovery (isolated)', () => {
+  beforeEach(() => {
+    jest.resetModules();
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('recovers after network outage - full flow', async () => {
+    // ---- In-memory DB ----
+    const txStore: any[] = [];
+    const repoMock = {
+      transactions: {
+        add: jest.fn(async (tx: any) => {
+          txStore.push({ ...tx });
+        }),
+        filter: jest.fn((fn: (tx: any) => boolean) => ({
+          toArray: jest.fn(async () => txStore.filter(fn))
+        })),
+        where: jest.fn((query: any) => ({
+          first: jest.fn(async () => txStore.find(tx => tx.id === query.id)),
+          modify: jest.fn(async (fn: (tx: any) => void) => {
+            const tx = txStore.find(t => t.id === query.id);
+            if (tx) fn(tx);
+          })
+        }))
+      }
+    };
+
+    // ---- Network toggle ----
+    let networkUp = true;
+    const mockSyncState = jest.fn(async () => {
+      if (!networkUp) throw new Error('Network unreachable');
+      return { blockNum: () => 42 };
+    });
+
+    const mockNewTransaction = jest.fn(async () => new Uint8Array([10, 20, 30]));
+
+    jest.doMock('lib/miden/repo', () => repoMock);
+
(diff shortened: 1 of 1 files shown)


WiktorStarczewski:
> LGTM
> 
> One thing we can also do that if we get that commitment mismatch error for public accounts we can just import account again i.e
> 
> delete the account from indexeddb and do `client.importAccountById` again I don't think we do this

But for public account the sync state loop will take care of it automagically no?

PR #151, opened 2026-03-06
Migrate to React SDK and v14 Miden Client

WiktorStarczewski:
Closes #152

After 0.14 client release:
TODO: Rewire localnet to testnet in the wallet (note this is now a 1-liner)
TODO: Rewire client
TODO: Rewire react-sdk

PR #154, opened 2026-03-10
feat: centralise network config and add localnet support

WiktorStarczewski:
## Summary

- Introduces a `DEFAULT_NETWORK` constant so the entire app can switch networks by changing a single value (currently set to `localnet`)
- Adds localnet note transport layer endpoint (`localhost:57292`)
- Graceful transaction recovery after network outages — AutoSync resumes cleanly and in-progress transactions are retried instead of silently dropped
- Simplifies network selection UI (removes unnecessary grouping)
- Widens `host_permissions` to `http://localhost/*` for local dev flexibility

## Test plan

- [ ] Verify wallet connects to local node on `localhost:57291`
- [ ] Verify note transport layer connects on `localhost:57292`
- [ ] Verify prover endpoint resolves on `localhost:50051`
- [ ] Test network switching in settings
- [ ] Confirm transaction recovery works after simulated network drop

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -15,8 +15,7 @@
 
   "permissions": ["storage", "clipboardWrite", "unlimitedStorage", "activeTab", "tabs", "notifications", "alarms"],
   "host_permissions": [
-    "http://localhost:4180/",
-    "http://localhost:3000/",
+    "http://localhost/*",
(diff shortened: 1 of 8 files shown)


WiktorStarczewski:
Included in https://github.com/0xMiden/wallet/pull/151, closing.

PR #159, opened 2026-03-16
Adding test for NTL reliability

0xnullifier:
In regard to #26, while the wallet does not do something special for the integration, it's a good idea to add a stress test this sends 100 privates notes through the NTL from the UI and checks it the notes are received in the UI

The test pass for me, but we can make the test more detailed somehow


WiktorStarczewski:
Superseded by the new test harness. Closing.

PR #162, opened 2026-03-17
feat: desktop notifications for new notes via service worker

WiktorStarczewski:
## Summary
- Service worker handles sync and background note checking for the extension
- Desktop notifications (Web Notifications API with chrome.notifications fallback) when new notes arrive while popup is closed
- Notification click opens fullscreen tab with instant note display from cached data
- Frontend sends SyncRequest every 3s to keep service worker alive; SW broadcasts SyncCompleted on each cycle

## Test plan
- [ ] Open popup, send note from another wallet — note appears within ~3s
- [ ] Close popup, send note — desktop notification appears within ~60s
- [ ] Click notification — fullscreen tab opens with notes displayed instantly
- [ ] Verify mobile is unaffected (all changes gated with `isExtension()`)

### DIFF
--- a/public/_locales/de/messages.json
+++ b/public/_locales/de/messages.json
@@ -2579,5 +2579,37 @@
   "type": {
     "message": "Typ",
     "englishSource": "Type"
+  },
+  "txIdLabel": {
+    "message": "Sende-ID",
+    "englishSource": "Tx ID"
+  },
+  "notesSection": {
+    "message": "Notes",
+    "englishSource": "Notes"
+  },
+  "created": {
+    "message": "Erstellt",
+    "englishSource": "Created"
+  },
+  "confirmed": {
+    "message": "Bestätigt",
+    "englishSource": "Confirmed"
+  },
+  "date": {
+    "message": "Datum",
+    "englishSource": "Date"
+  },
+  "details": {
+    "message": "Einzelheiten",
+    "englishSource": "Details"
+  },
+  "addANote": {
+    "message": "Fügen Sie ein note hinzu",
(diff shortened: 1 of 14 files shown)


src/lib/miden/back/sync-manager.ts
```diff
@@ -0,0 +1,236 @@
...
+        const { base } = await fetchTokenMetadata(note.faucetId);
+        note.metadata = {
+          decimals: base.decimals,
+          symbol: base.symbol,
+          name: base.name,
+          thumbnailUri: base.thumbnailUri
+        };
+      } catch {
+        // Leave metadata undefined — frontend will handle
+      }
```
0xnullifier: We also cache metadata right so wouldn't it be better to see if the the metadata is already cached or not, we can also change the `fetchTokenMetadata` to check for cached metadata because currently it always queries the rpc

src/lib/miden/front/claimable-notes.ts
```diff
@@ -144,33 +174,66 @@ export function useClaimableNotes(publicAddress: string, enabled: boolean = true
...
-    const parsedNotes = parseNotes(rawNotes, notesBeingClaimed);
+            // Warm up local client in background for subsequent fetches
+            withWasmClientLock(async () => {
+              const client = await getMidenClient();
+              await client.syncState();
+            })
+              .then(() => {
+                localClientReady.current = true;
+              })
+              .catch(() => {});
```
0xnullifier: Hmm why do we need to "warm up" the client seems a little unnecessary no?

PR #163, opened 2026-03-17
feat: service worker as single backend for extension reads

WiktorStarczewski:
## Summary
- SW broadcasts notes + vault assets via `SyncCompleted` after each sync cycle — frontend reads from Zustand only, no WASM client needed for reads
- `useClaimableNotes` dispatches to extension hook (Zustand) or local hook (WASM) based on `isExtension()`
- `useAllBalances` skips WASM polling on extension — balances arrive via broadcast
- Skip eager WASM client init on extension (lazy-init on first write operation)
- Cross-tab claim coordination via `NoteClaimStarted` broadcast
- New `updateBalancesFromSyncData` utility converts `SerializedVaultAsset[]` to `TokenBalanceData[]`

## Test plan
- [ ] Open popup, send note from another wallet — note appears within ~3s
- [ ] Open fullscreen tab, send note — note appears within ~3s
- [ ] Close popup, send note — background notification within ~60s, click opens instant display
- [ ] Balances update in popup within ~3s after sending tokens
- [ ] Switch accounts — data refreshes for new account
- [ ] Open popup + fullscreen — claim note in one, other shows "being claimed" instantly
- [ ] Mobile unaffected — verify AutoSync still works
- [ ] `yarn lint && yarn build && yarn test` — all pass (901 tests)


src/lib/miden/back/sync-manager.ts
```diff
@@ -34,96 +34,125 @@ export async function doSync(): Promise<void> {
...
+      await Promise.all(
+        parsedNotes.map(async note => {
+          try {
+            const { base } = await fetchTokenMetadata(note.faucetId);
+            note.metadata = {
+              decimals: base.decimals,
+              symbol: base.symbol,
+              name: base.name,
+              thumbnailUri: base.thumbnailUri
+            };
```
0xnullifier: I think it's cleaner to use a `as` cast here it will work no?

PR #176, opened 2026-04-01
feat: Chrome side panel mode with popup toggle

WiktorStarczewski:
## Summary

- Add Chrome Side Panel as an alternative display mode (popup remains the default)
- Users can toggle between popup and side panel via the maximize/minimize icon in the header
- Preference is persisted across sessions via `chrome.storage.local`
- Add pin extension prompt on first install
- Color-coded Send (blue) and Receive (green) action buttons on home page
- Update `@miden-sdk/miden-sdk` to 0.13.3

## Details

**Side panel infrastructure:**
- New `WindowType.SidePanel` enum variant and `compact` boolean (`popup || sidePanel`) in `useAppEnv()`
- New `sidepanel.html` (full-height layout) and `sidepanel.tsx` entry point
- Manifest: `side_panel` config, `sidePanel` permission, Chrome min version bumped to 114
- Webpack: sidepanel entry point and HTML template added

**Mode toggle:**
- Maximize icon in popup → opens side panel, disables popup on icon click
- Minimize icon in side panel → closes panel, restores popup on icon click
- Background script restores preference on startup via `chrome.action.setPopup()` and `chrome.sidePanel.setPanelBehavior()`

**Layout updates:**
- Side panel uses `height: 100%; width: 100%` (not fixed 360×600px)
- All container sizing components updated: PageLayout, TabLayout, SimplePageLayout, Receive, TokenDetail, SendManager, ConsumingNote

**Other:**
- `popup` checks migrated to `compact` across 10+ components for shared narrow-width behavior
- Pin extension prompt shown once after fresh install
- Onboarding flow constrained to 420px max-width

## Test plan

- [ ] Click extension icon → popup opens (default)
- [ ] Click maximize icon in popup → side panel opens, popup closes
- [ ] Click extension icon while in side panel mode → side panel opens (not popup)
- [ ] Click minimize icon in side panel → panel closes
- [ ] Click extension icon after minimizing → popup opens
- [ ] Reload extension → mode preference persists
- [ ] Fresh install → fullpage tab opens with pin prompt
- [ ] Send/Receive buttons on home page show blue/green colors
- [ ] All screens render correctly in side panel (full height, no overflow)

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,5 +1,15 @@
 /* eslint-disable */
 
+// IMPORTANT: Register onInstalled BEFORE importScripts so it fires synchronously.
+// Webpack's async module loading in background.js can delay listener registration,
+// causing Chrome MV3 to miss the install event.
+chrome.runtime.onInstalled.addListener(function (details) {
+  if (details.reason === 'install') {
+    chrome.storage.local.set({ 'fresh_install': true });
+    chrome.tabs.create({ url: chrome.runtime.getURL('fullpage.html') });
+  }
+});
+
 try {
   const window = globalThis;
   // This is the file produced by webpack
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,14 @@
 
 ### Features
 
+* [FEATURE][extension] Chrome Side Panel mode with popup toggle. Users can switch between popup (default) and side panel via the maximize/minimize icon in the header. Preference persists across sessions. (#176)
+* [FEATURE][extension] Pin extension prompt shown once after fresh install, guiding users to pin the extension to the toolbar. (#176)
+* [FEATURE][all] Color-coded Send (blue) and Receive (green) action buttons on the home page, matching the token detail page. (#176)
+
+### Fixes
+
+* [FIX][extension] Fixed `onInstalled` event handler not firing on Chrome MV3 due to webpack async module loading delaying listener registration. Handler moved to `sw.js` for synchronous registration. (#176)
+* [FIX][all] Fixed `ConsumingNote` page using raw UA sniffing instead of `isMobile()` platform detection. (#176)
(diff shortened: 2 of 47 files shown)


0xnullifier:
We will merge the v14 soon right? so wouldn't it be better to base this off v14

WiktorStarczewski:
> We will merge the v14 soon right? so wouldn't it be better to base this off v14

There shouldnt be that many conflicts right, since its presentation layer?

public/sw.js
```diff
@@ -1,5 +1,15 @@
...
+// IMPORTANT: Register onInstalled BEFORE importScripts so it fires synchronously.
+// Webpack's async module loading in background.js can delay listener registration,
+// causing Chrome MV3 to miss the install event.
+chrome.runtime.onInstalled.addListener(function (details) {
+  if (details.reason === 'install') {
+    chrome.storage.local.set({ 'fresh_install': true });
+    chrome.tabs.create({ url: chrome.runtime.getURL('fullpage.html') });
+  }
+});
+
```
0xnullifier, line 12: The onboarding UI does not look good in full page so we should modify that if we want to do this

I think the previous approach was fine i.e onboarding in the popup or we could open the sidebar instead

0xnullifier (commented):
I think a close button in sidebar mode would be helpful as well

public/sw.js
WiktorStarczewski: The problem is that you can open neither (neither the popup, nor the sidebar) automatically, so after installing the wallet the user is stuck in a "nothing happened" phase, where literally nothing happens. I've remedied some of that I think by adding the tooltip asking to pin the extension, but I feel like that's kind of not enough. I think most users would expect "something" to happen when they install a wallet, as its really (installing + onboarding) one mental flow, not two, as it is when we dont do full page onboarding. Happy to hear counterarguments though. When you say "the onboarding UI does not look good", wdym exactly?

PR #184, opened 2026-04-14
fix: dedup consume transactions across Completed state (#171)

### LINKED ISSUE #171: Multiple consuming note notifications for a single transaction
Multiple consuming notes notification in wallet history in one transaction

Steps to Reproduce
- Open the Miden extension wallet
- Initiate and complete any transaction
- Observe on the recipient address
- 1 note has multiple consuming notification on the wallet history

https://github.com/user-attachments/assets/f22bf078-03b5-49d5-898a-62ff77973917

WiktorStarczewski:
Closes #171.

## Summary
- Auto-consume was creating a fresh consume transaction every 5s SWR poll, producing up to 150 duplicate `Completed` rows for a single received note.
- Root cause: `initiateConsumeTransaction` deduped only against `Queued` / `GeneratingTransaction` transactions. After the first consume completed, the WASM client's `getConsumableNotes()` briefly kept returning the note (chain-sync lag), so each poll inserted a new row.

## Changes
- **`src/lib/miden/activity/transactions.ts`** — dedup now checks all non-`Failed` consume txs for the same `(noteId, accountId)`, including `Completed`. `Failed` is excluded so retries still work.
- **`src/lib/miden/repo.ts`** — Dexie schema bumped to `1.2` with a new `noteId` index so the dedup lookup is O(log n) instead of a full-table scan. Dexie reindexes existing rows automatically on open — no data migration.
- **`src/lib/store/hooks/useIntercomSync.ts`** — the 3s sync poll no longer blanket-clears `extensionClaimingNoteIds`, which was defeating Explore's `isBeingClaimed` gate. It now removes only IDs whose notes are no longer consumable in the SW's sync data.
- **`src/lib/store/{index,types}.ts`** — new `removeExtensionClaimingNoteIds(ids)` action backing the surgical removal above.

## Tests
- Regression cases in `transactions.test.ts` and `transactions.extended.test.ts`: Completed dedup, Failed retry, cross-account non-dedup.
- `store/index.test.ts`: coverage for `removeExtensionClaimingNoteIds` (partial removal + no-op paths).
- Full suite: **1647 passing, 121 suites**. Lint, prettier, and tsc clean.

## Test plan
- [x] `yarn test` — all suites pass
- [x] `yarn lint` — no warnings
- [x] `yarn format` — clean
- [x] `yarn tsc --noEmit` — clean
- [ ] Manual verification on an account with an unconsumed note: confirm the wallet produces exactly one `Consumed` history entry after receive.

## Notes for reviewers
- I traced the Rust-side question of *why* the SDK's `getConsumableNotes` re-returns a locally-consumed note (filter is correct, `apply_transaction` correctly flips `Committed` → `ProcessingAuthenticated`, and a repeat consume should `Err(NoteNotConsumable)`). The 150 `Completed` rows observed in the bug report can't be explained by that state machine alone — something is either resetting state back to `Committed` or the wallet is marking Completed despite a Rust failure. Without a live repro I can't pin that down; the JS dedup here is a defensive fix that works regardless of the root cause and is worth landing on its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 * [FEATURE][all] Migrated backend from `WasmWebClient` to the new `MidenClient` TypeScript API. All service-worker WASM access now goes through `MidenClientInterface` wrapping the high-level `MidenClient` surface.
 * [FEATURE][all] Migrated frontend to `@miden-sdk/react` hooks (`useMiden`, `useSyncState`, `useAccount`, etc.), replacing manual sync and balance-polling logic.
 
+### Fixes
+
+* [FIX][all] Fixed duplicate consume-transaction entries in wallet history when receiving a single note. `initiateConsumeTransaction` now dedups against all non-`Failed` consume txs for the same note (including `Completed`), preventing auto-consume from re-enqueueing while `getConsumableNotes` is still returning the note during chain-sync lag. Also replaced the sync poll's blanket clear of `extensionClaimingNoteIds` with a surgical remove so Explore's `isBeingClaimed` gate works correctly. (#184)
+
 ---
 
 ## 1.13.3 (2026-03-19)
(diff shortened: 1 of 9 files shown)


0xnullifier (approved):
LGTM

PR #188, opened 2026-04-15
test: add randomized stress suite for real-world send/claim patterns

WiktorStarczewski:
## Summary

Adds a second Chrome E2E test suite alongside the existing blockchain-ops tests, focused on simulating real-world usage over a long run: two wallets randomly sending public/private notes back and forth, periodically claiming, with perturbations mixed in (lock/unlock, page reload, concurrent sends, idle periods). Parametrized by note count so it can be dialled from a ~15 min smoke to a multi-hour soak.

Asserts **strict balance conservation** after the final drain phase — any lost note means a real bug, not acceptable flakiness.

## What's in

```
playwright.stress.config.ts                # separate config, no per-test timeout
playwright/e2e/stress/stress.spec.ts       # setup, drive, drain, verify
playwright/e2e/stress/stress-driver.ts     # seeded RNG, op loop, perturbations
```

Plus small supporting changes:
- `WalletPage.sendTokens` gains an optional `tokenSymbol` param. MIDEN ships as an auto-tracked 0-balance entry in the wallet's `useAllBalances` list; `.first()` picks it and the send validates against 0. The new param lets the driver target `"TST"` (the stress suite's custom faucet symbol). iOS POM mirrors the signature for interface compat.
- `EventCategory` union adds `'stress_op'` so driver events are typed.
- 6 new yarn scripts: `test:e2e:stress[:build|:run|:testnet|:devnet|:localhost]`.

## Parameters (all env vars)

| Var | Default | Notes |
|---|---|---|
| `STRESS_NUM_NOTES` | 20 | stop condition; ≈15 min on devnet |
| `STRESS_INITIAL_MINTS` | 3 | 3000 TST starting per wallet |
| `STRESS_AMOUNT_MIN` / `STRESS_AMOUNT_MAX` | 1 / 10 | display units |
| `STRESS_DELAY_MIN_MS` / `STRESS_DELAY_MAX_MS` | 3000 / 10000 | inter-send gap |
| `STRESS_PRIVATE_RATIO` | 0.5 | public vs private mix |
| `STRESS_CLAIM_AFTER_SEND_PROB` | 0.5 | receiver claims immediately |
| `STRESS_IDLE_EVERY` / `_MIN_MS` / `_MAX_MS` | 10 / 30000 / 60000 | user-stepped-away simulation |
| `STRESS_LOCK_EVERY` | 15 | lock/unlock cycle frequency |
| `STRESS_RELOAD_EVERY` | 20 | page reload cycle frequency |
| `STRESS_CONCURRENT_PROB` | 0.15 | both wallets send at once |
| `STRESS_SEED` | `Date.now()` | reproducibility |
| `STRESS_CONSERVATION_STRICT` | true | fail on any balance delta |

Set any of `STRESS_LOCK_EVERY`, `STRESS_RELOAD_EVERY`, `STRESS_IDLE_EVERY`, `STRESS_CONCURRENT_PROB` to 0 to disable that perturbation. Set `STRESS_NUM_NOTES` to 500+ for multi-hour soak runs.

## Output artifacts

Every run writes to the timeline's output dir:
- `stress-summary.json` — seed, options, initial/final balances, per-op latency percentiles, perturbation counts, conservation-held flag
- `stress-operations.csv` — per-op record (sender, receiver, isPrivate, amount, sendMs, status, concurrent, perturbation, error)
- `timeline.ndjson` — full event stream (same schema as the existing E2E suite)

Playwright traces land under `test-results-stress/` (kept separate from the regular `test-results/` so nothing clobbers across runs).

## Verification

- `yarn lint`, `yarn format`, `yarn test` all green (1656 tests, coverage 95%+)
- Smoke run on devnet (`STRESS_NUM_NOTES=2`, no perturbations): 2/2 sends completed, 0 failed, conservation held (delta=0), 6.9 min total including setup + 5-cycle drain
- Separately kicked off a 100-note run to exercise the perturbation paths end-to-end — results will be captured and investigated if anything goes sideways

### Default run (~15 min on devnet)

```bash
yarn test:e2e:stress:devnet
```

20 notes, all perturbations on, new random seed each time.

### Quick sanity check after a wallet change (~7 min)

```bash
STRESS_NUM_NOTES=5 \
STRESS_LOCK_EVERY=0 STRESS_RELOAD_EVERY=0 STRESS_IDLE_EVERY=0 STRESS_CONCURRENT_PROB=0 \
yarn test:e2e:stress:devnet
```

Base loop only, no perturbations — fastest path to "does a random send/claim actually work end-to-end against my build."

### Overnight soak (~100 min)

```bash
STRESS_NUM_NOTES=200 yarn test:e2e:stress:devnet
```

### Reproduce a specific failure

Every run logs `seed=<number>` at the top of stdout and inside `stress-summary.json`. Re-run with that seed to replay the exact same sender/receiver/amount/perturbation sequence:

```bash
STRESS_SEED=2433243116 STRESS_NUM_NOTES=20 yarn test:e2e:stress:devnet
```

### Focus on a single perturbation

Say you just fixed a bug in the reload path and want dense reload coverage without the rest:

```bash
STRESS_NUM_NOTES=30 \
STRESS_RELOAD_EVERY=3 \
STRESS_LOCK_EVERY=0 STRESS_IDLE_EVERY=0 STRESS_CONCURRENT_PROB=0 \
yarn test:e2e:stress:devnet
```

### Skip the build step when iterating

```bash
# Rebuild once:
yarn test:e2e:stress:build
# Then re-run the test only (no rebuild):
STRESS_NUM_NOTES=10 yarn test:e2e:stress:run
```

### Switching networks

```bash
yarn test:e2e:stress:testnet       # testnet
yarn test:e2e:stress:localhost     # expects a local node on :57291
```

## Test plan

- [ ] `yarn test:e2e:stress:devnet` runs to completion on devnet with default params
- [ ] Run with `STRESS_NUM_NOTES=5 STRESS_LOCK_EVERY=0 STRESS_RELOAD_EVERY=0 STRESS_IDLE_EVERY=0 STRESS_CONCURRENT_PROB=0` as a fast smoke — confirms base loop
- [ ] `stress-summary.json` and `stress-operations.csv` are present in the run's artifact dir and parse correctly
- [ ] Strict conservation assertion fires if a note is artificially lost (can smoke-test by forcing an exception inside `claimAllNotes` and checking the assertion message)


src/components/TransactionProgressModal.tsx
```diff
@@ -31,22 +31,25 @@ export const TransactionProgressModal: FC = () => {
   // Track if we're actively processing (started when modal opens, continues even when hidden)
   const [isProcessing, setIsProcessing] = useState(false);
 
-  // On extension: check for uncompleted send transactions on mount and auto-open modal
+  // On extension: if there are uncompleted send transactions on mount, nudge
```
0xnullifier: What about mobile then? to me it seems the uncompleted transactions on mobile are not handled

0xnullifier: Because on mobile the modal actually starts the transaction loop instead of the sw so idk if this addresses that case

0xnullifier (commented):
I think the mobile case for uncompleted transactions that we used to do `openModal` is not addressed

PR #214, opened 2026-04-25
Service worker init race causing stuck transactions (#212)

### LINKED ISSUE #212: Transaction stuck in “sending” state for extended period
# Description

A manually initiated transaction became stuck in the “sending” state for more than 30 minutes. Although it eventually appeared to go through, it was not recorded in the transaction history as either a successful or failed transaction.

Additionally, the wallet balance was not deducted, creating uncertainty about whether the transaction was actually processed.

_________________________________________________________________________________________________________________________

## Summary

A manually-initiated transaction has been stuck in "sending" (`GeneratingTransaction`) state for 20+ minutes because the Chrome extension's service worker fails to complete initialization on a cold-start after SW idle-termination. Both `Actions.init()` and `doSync()` access `Vault.isExist()` before the bundled module initializer `init_vault()` has finished running — `Vault === undefined`, producing `TypeError: Cannot read properties of undefined (reading 'isExist')` at two independent call sites. Because `Actions.init()` threw, `inited(vaultExist)` was **likely** never reached (line 62 of `actions.ts` is unreachable once line 60 throws), so the backend's `inited` flag **would therefore stay false** — every `withInited(...)` wrapper (the standard backend guard for transaction processing) blocks on the pre-init state indefinitely, **consistent with** the observed stuck transaction. The root cause is a module-initialization race introduced by the Vite SW build: the `sw-patches` plugin strips top-level `await` via a blanket regex and then re-sequences inits through a `__initsReady` barrier, but (a) `main.ts::start()` calls `Actions.init()` outside that barrier, and (b) the barrier is explicitly skipped for SYNC_REQUEST handlers. Reproduced on wallet **v1.14.3**, commit **`284e99e`** (current `main` HEAD), with `@miden-sdk/miden-sdk v0.14.2` + `@miden-sdk/react v0.14.2` + `@miden-sdk/vite-plugin v0.14.2` (verified from `package.json`). Live reproduction not attempted — analysis based on user-provided console logs + code inspection only.

## Symptoms

Exact console output (user-provided, verbatim):

```
[GeneratingTransaction] State: Object
chunks/run-checks.heyQTwF9.js:29 Uncaught (in promise) Error: Could not establish connection. Receiving end does not exist.
transactions.ts:48 [init_transactions] 1: init_dexie_min()
transactions.ts:54 [init_transactions] 2: init_repo()
transactions.ts:54 [init_transactions] 3: init_platform()
transactions.ts:59 [init_transactions] 4: await init_store()...
main.ts:49 Miden background script started
actions.ts:86 [Actions.init] Starting...
transactions.ts:59 [init_transactions] 4: init_store() done
transactions.ts:61 [init_transactions] 5: init_logger()
transactions.ts:67 [init_transactions] 6: init_helpers()
transactions.ts:69 [init_transactions] 7: await init_miden_client()...
transactions.ts:69 [init_transactions] 7: init_miden_client() done
transactions.ts:75 [init_transactions] 8: await init_notes()...
transactions.ts:75 [init_transactions] 8: init_notes() done
actions.ts:87 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'isExist')
    at init (actions.ts:87:19)
    at start (main.ts:51:10)
    at background-entry.ts:12:50
main.ts:75 [processRequest] type: SYNC_REQUEST
sync-manager.ts:198 [SyncManager] Sync error: TypeError: Cannot read properties of undefined (reading 'isExist')
    at doSync (sync-manager.ts:57:20)
    at processRequest (main.ts:77:61)
    at server.ts:151:8
    at IntercomServer.processMessage (server.ts:183:11)
    at IntercomServer.handleMessage (server.ts:146:11)
actions.ts:86 [Actions.init] Starting...
```

**User-visible:** third manually-initiated transaction has remained in "sending" state for 20+ minutes with no advance. **Persistence scope:** by mechanism, the backend's `inited` flag cannot transition to `true` without a successful `Actions.init()` call — so in the absence of a fresh SW cold-start where the race resolves favorably, the state is expected to persist.

## Reproduction steps

1. Running wallet `v1.14.3` on Chrome (MV3) built from commit `284e99e`.
2. Wallet is unlocked and has an established vault.
3. User initiates transactions manually via the UI (any type — send, consume).
4. At some point, Chrome terminates the MV3 service worker after its idle window (~30s of inactivity).
5. On the next event (user action, sync alarm, or frontend `SYNC_REQUEST`), Chrome cold-starts the SW by re-executing `background.js`.
6. The SW cold-start triggers `start()` via `background-entry.ts:16` → `./background` top-level execution. `start()` is **not** wrapped in the plugin's `__initsReady` barrier; it calls `Actions.init()` directly as its first act.
7. `Actions.init()` executes `await Vault.isExist()` while `init_vault()` is still pending — `Vault === undefined` → `TypeError`.
8. Frontend dispatches `SYNC_REQUEST` → `processRequest` handles it → plugin-patched guard **skips** `await __initsReady` for SYNC_REQUEST → `doSync()` runs `await Vault.isExist()` → same `TypeError`.
9. Both errors are "Uncaught (in promise)" — swallowed by the runtime with no UI feedback. Any transaction already in `GeneratingTransaction` remains stuck because the `withInited(...)` guard on its processing pipeline never unblocks.

- **Expected:** SW completes initialization deterministically; `Actions.init()` calls `inited(vaultExist)`; backend transitions to `inited: true`; `withInited`-guarded transaction processing proceeds; the in-flight 3rd transaction either completes or fails explicitly.
- **Actual:** SW init throws on `Vault.isExist` against `undefined`; `inited` never fires; every `withInited(...)` wrapper blocks forever; transaction is indefinitely stuck with no user-visible error.
- **Reproduction rate:** likely high on any MV3 SW cold-start after the wallet is already unlocked with an in-flight or queued transaction. The race depends on whether `init_vault()` resolves before `start()`'s `await Actions.init()` runs, which depends on bundle-specific TLA-stripping patterns — fragile to dependency-tree changes.
- **Live reproduction status:** not attempted — analysis based on user-provided console logs + code inspection only.

## Root cause (3 locations — 1 primary throw site + 1 sibling throw site + 1 configuration source)

**[`src/lib/miden/back/actions.ts#L58-L64`](../blob/main/src/lib/miden/back/actions.ts#L58-L64)** — primary throw site

```ts
export async function init() {
  console.log('[Actions.init] Starting...');
  const vaultExist = await Vault.isExist();   // Vault === undefined during cold-start race
  console.log('[Actions.init] Vault exists:', vaultExist);
  inited(vaultExist);
  console.log('[Actions.init] Called inited()');
}
```

`Vault` is imported from `./vault` (actions.ts:16). In the Vite SW bundle, `Vault` is assigned inside `init_vault()`'s factory body; if `init_vault()` hasn't run, the binding reads as `undefined`. When `start()` calls `Actions.init()` before `init_vault()` resolves, this line throws. The `await Actions.init()` in `main.ts::start()` rejects, so the rest of `start()` (including `frontStore = store.map(toFront)` and the `StateUpdated` broadcast at main.ts:28-32) never executes — the backend never signals readiness to the frontend.

**[`src/lib/miden/back/sync-manager.ts#L19-L36`](../blob/main/src/lib/miden/back/sync-manager.ts#L19-L36)** — sibling throw site, same race

```ts
export async function doSync(): Promise<void> {
  if (isSyncing) return;
  isSyncing = true;
  try {
    const exists = await Vault.isExist();   // same undefined-Vault pattern
    ...
```

`doSync()` is called from `processRequest` on `SYNC_REQUEST`. Per the plugin-patched guard, SYNC_REQUEST bypasses `await __initsReady`, so `doSync()` races module init independently of `start()`.

**[`vite.background.config.ts#L77`, `#L161-L166`, `#L201`](../blob/main/vite.background.config.ts#L77)** — configuration source

```js
// Line 77 — strips ALL top-level await:
chunk.code = chunk.code.replace(/^await /gm, '/* tla-stripped */ ');
...
// Lines 161-166 — init_actions excluded from the re-sequencing barrier:
const coreInits = uniqueInits.filter(c =>
  !c.includes('init_actions') && !c.includes('init_transaction_processor')
);
...
// Line 201 — barrier bypassed for SYNC_REQUEST:
'$&\n  if (req?.type !== "GET_STATE_REQUEST" && req?.type !== "SYNC_REQUEST") { await __initsReady; }'
```

TLA stripping is necessary for ESM-SW compatibility, but the regex-based re-sequencing is fragile: `init_actions` is explicitly excluded (the comment at lines 157-160 explains `init_actions → init_dapp → init_activity` hangs in SW context), and SYNC_REQUEST is excluded from the barrier. Both exclusions let unresolved module bindings leak into runtime paths.

## Receipts (evidence chain)

| Signal | Source | What it proves |
|---|---|---|
| Exact `Vault.isExist()` call in source | [`src/lib/miden/back/actions.ts#L60`](../blob/main/src/lib/miden/back/actions.ts#L60) | Primary throw site; `Vault` is accessed unguarded at init-time |
| Second `Vault.isExist()` call in source | [`src/lib/miden/back/sync-manager.ts#L25`](../blob/main/src/lib/miden/back/sync-manager.ts#L25) | Same error signature at second unrelated call site — confirms module-wide race, not a site-specific bug |
| Author comment flagging this exact init pattern | [`src/lib/miden/back/actions.ts#L43-L46`](../blob/main/src/lib/miden/back/actions.ts#L43-L46): *"in the Vite SW build, module-scope init (init_actions) may not complete because it transitively depends on dapp.ts which imports frontend modules that hang in SW context. Making queues lazy ensures they're available on first use regardless of whether init_actions completed."* | Codebase already acknowledged this class of bug and partially worked around it (lazified `_dappQueue`/`_unlockQueue`) — but did NOT lazify the `Vault` reference in `init()` |
| Second author comment flagging TLA stripping | [`src/lib/miden/back/main.ts#L15-L17`](../blob/main/src/lib/miden/back/main.ts#L15-L17): *"with Vite's TLA stripping, `store` may not be initialized at module scope evaluation time"* | Independent author acknowledgment of the same class of bug on a sibling module; the TLA-stripping problem is known and partially mitigated |
| Literal TLA-stripping regex | [`vite.background.config.ts#L77`](../blob/main/vite.background.config.ts#L77) — `chunk.code.replace(/^await /gm, '/* tla-stripped */ ')` | The mechanism that breaks implicit init-ordering across the entire SW bundle |
| `init_actions` excluded from the `__initsReady` barrier | [`vite.background.config.ts#L161-L166`](../blob/main/vite.background.config.ts#L161-L166) + comment at L156-L160 (*"init_actions → init_dapp → init_activity never resolves..."*) | No ordered barrier sequences `init_vault` (transitive dep of `init_actions`) before `Actions.init()` runs via the plugin's generated `await init()` line |
| SYNC_REQUEST bypasses the barrier | [`vite.background.config.ts#L201`](../blob/main/vite.background.config.ts#L201) + comment at L197-L198 (plugin assumes SYNC_REQUEST is *"handled early via getFrontState's Idle fallback"*) | Second throw site (`sync-manager.ts:25`) has its own race path, independent of the `start()` race — the plugin's assumption about SYNC_REQUEST being pre-init-safe is empirically wrong |
| Console log ordering | User-provided log: `[init_transactions] 4: await init_store()...` fires, then `main.ts:49 Miden background script started` + `actions.ts:86 [Actions.init] Starting...` fire **before** `[init_transactions] 4: init_store() done` | Direct empirical evidence that `start()` and `Actions.init()` execute while the init chain is still in flight — the race is observably real in the user's runtime |
| Identical error signature at both call sites | User log: `TypeError: Cannot read properties of undefined (reading 'isExist')` from both `init (actions.ts:87:19)` and `doSync (sync-manager.ts:57:20)` | Both sites read `Vault` as undefined independently; consistent with module-init race, not a per-call-site bug |

## Suggested direction

The workaround pattern already in-repo is the model: lazy accessors. The fix is to apply it to `Vault`. Either (a) introduce a `getVault()` lazy accessor that awaits `init_vault()` (idempotent via `__esmMin`) before returning the class, used everywhere that reads `Vault` at module-use time in backend code (actions.ts, sync-manager.ts, dapp.ts, store.ts); or (b) add explicit `await __initsReady` guards at the top of `Actions.init()` and `doSync()` — but this requires lifting `init_actions` exclusion from `coreInits` first, which the plugin comment warns against because `init_actions → init_dapp → init_activity` hangs in SW context, so option (a) is strictly safer and mirrors the existing [`actions.ts#L47-L56`](../blob/main/src/lib/miden/back/actions.ts#L47-L56) lazy-queue pattern. A targeted third mitigation: remove SYNC_REQUEST from the bypass list in [`vite.background.config.ts#L201`](../blob/main/vite.background.config.ts#L201) — the plugin comment assumes SYNC_REQUEST is safe pre-init, but the current log empirically disproves that assumption. For a long-term structural fix, the regex-based re-sequencing approach in `sw-patches` is fragile by design; replacing it with an explicit init manifest maintained in source (naming modules that must be awaited at SW bootstrap) would remove the hidden coupling between bundler output patterns and SW correctness.

____________________________________________________________________________________________________________________

## Out-of-scope note (not part of the main report):

 The log shows chunks/run-checks.heyQTwF9.js:29 Uncaught (in promise) Error: Could not establish connection. Receiving end does not exist. — this is a secondary signal of SW lifecycle instability (content script / popup messaging an SW that was terminated), related to but distinct from the primary stuck-transaction mechanism.

https://github.com/user-attachments/assets/1cef2688-3012-4b16-b60f-55426df02c09

verseon0980:
## Summary
Fixes #212 - Service worker initialization race condition that causes transactions to remain stuck in "sending" state indefinitely after Chrome MV3 service worker cold-start.

## Problem
When Chrome terminates the MV3 service worker after idle timeout (~30s), the next cold-start triggers a module initialization race:

1. `Actions.init()` and `doSync()` both call `Vault.isExist()` before `init_vault()` completes
2. `Vault` is `undefined` during the race → `TypeError: Cannot read properties of undefined (reading 'isExist')`
3. `Actions.init()` throws, so `inited(vaultExist)` never fires
4. Backend's `inited` flag stays `false`
5. All `withInited(...)` guards block forever
6. Any in-flight transaction is permanently stuck

**Root cause locations:**
- `src/lib/miden/back/actions.ts:60` - unguarded `Vault.isExist()` call
- `src/lib/miden/back/sync-manager.ts:69` - same unguarded access
- `vite.background.config.ts:201` - SYNC_REQUEST bypasses init barrier

## Solution
Applied the **lazy accessor pattern** already used in the same codebase (`_dappQueue`/`_unlockQueue` at `actions.ts:47-56`) to the `Vault` reference:

### Changes Made:
1.  **`sync-manager.ts`**: Added `getVault()` lazy accessor and replaced direct `Vault` calls with `await getVault()`
2. **`actions.ts`**: Applied the same lazy accessor pattern (if not already present)
3. **`vite.background.config.ts`**: Removed `SYNC_REQUEST` from the `__initsReady` bypass so sync requests properly wait for initialization

This ensures `Vault` is always initialized before use, regardless of service worker cold-start timing.

## Testing
- Code follows existing lazy-init pattern from the same file
- Fixes both identified throw sites (`actions.ts` + `sync-manager.ts`)
- Removes the incorrect SYNC_REQUEST bypass assumption

## Related
- Confirmed by user console logs + video in #212
- No existing PR addresses this issue
- Issue opened Apr 23, 2026 with full forensic analysis

### DIFF
--- a/src/lib/miden/back/actions.ts
+++ b/src/lib/miden/back/actions.ts
@@ -55,9 +55,20 @@ function getUnlockQueue() {
   return _unlockQueue;
 }
 
+// service worker cold-start race
+let _vault: typeof Vault | null = null;
+async function getVault() {
+  if (!_vault) {
+    await init_vault(); // init_vault completes before returning
+    _vault = Vault;
+  }
+  return _vault;
+}
+
 export async function init() {
   console.log('[Actions.init] Starting...');
-  const vaultExist = await Vault.isExist();
+  const vault = await getVault(); // wait for vault initialization
+  const vaultExist = await vault.isExist();
   console.log('[Actions.init] Vault exists:', vaultExist);
   inited(vaultExist);
   console.log('[Actions.init] Called inited()');
--- a/src/lib/miden/back/sync-manager.ts
+++ b/src/lib/miden/back/sync-manager.ts
@@ -10,6 +10,7 @@ import { mergeAndPersistSeenNoteIds } from './note-checker-storage';
 import { Vault } from './vault';
 import { getBech32AddressFromAccountId } from '../sdk/helpers';
 import { getMidenClient, withWasmClientLock } from '../sdk/miden-client';
+import { init_vault } from './vault';
 
 const ALARM_NAME = 'miden-sync';
 
@@ -39,6 +40,16 @@ let inFlight: Promise<void> | null = null;
 let consecutiveSyncFailures = 0;
 let syncBackoffUntilMs = 0;
 
+// Lazy Vault initialization to prevent service worker cold-start race
+let _vault: typeof Vault | null = null;
+async function getVault() {
+  if (!_vault) {
+    await init_vault();
+    _vault = Vault;
+  }
+  return _vault;
+}
+
 export function doSync(): Promise<void> {
   if (inFlight) return inFlight;
   // Circuit-breaker: short-circuit if recent syncs failed and we're waiting out
@@ -56,7 +67,8 @@ export function doSync(): Promise<void> {
 async function runSync(): Promise<void> {
   try {
     // Skip if wallet not set up
-    const exists = await Vault.isExist();
+    const vault = await getVault();
+    const exists = await vault.isExist();
     if (!exists) return;
 
     // [Lock 1] THE sync for the whole app. Bounded by SYNC_TIMEOUT_MS so it
@@ -89,7 +101,8 @@ async function runSync(): Promise<void> {
     }
 
     const intercom = getIntercom()!;
-    const accountPubKey = await Vault.getCurrentAccountPublicKey();
+    const vault2 = await getVault();
+    const accountPubKey = await vault2.getCurrentAccountPublicKey();
 
     if (accountPubKey) {
       // [Lock 2] Read notes + vault assets from warm WASM client
--- a/vite.background.config.ts
+++ b/vite.background.config.ts
@@ -196,11 +196,11 @@ export default defineConfig({
                 'async function processRequest'
               ].join('\n')
             );
-            // processRequest awaits inits for any request except GetStateRequest/SyncRequest
-            // (those are handled early via getFrontState's Idle fallback)
+            // processRequest awaits inits for any request except GET_STATE_REQUEST
+            // SYNC_REQUEST removed from bypass - it should now properly wait for init
             chunk.code = chunk.code.replace(
               /async function processRequest\(req[^)]*\)\s*\{/,
-              '$&\n  if (req?.type !== "GET_STATE_REQUEST" && req?.type !== "SYNC_REQUEST") { await __initsReady; }'
+              '$&\n  if (req?.type !== "GET_STATE_REQUEST") { await __initsReady; }'
             );
           }
         }


WiktorStarczewski:
Closing — this fix shipped via #224, which was merged a few minutes ago and closes #212.

#224 cherry-picked all three of your commits unchanged (`e20e0fa`, `ae2f97b`, `166f7541`) and added a follow-up commit that:

1. Extends the `getVault()` lazy accessor to a third cold-start site (`isDAppEnabled` in `actions.ts`) that wasn't covered here.
2. Fixes a TypeScript compile error this PR's `import { init_vault } from './vault'` would have hit — `init_vault` isn't a source-level export, it's auto-injected by the `sw-patches` Vite plugin, so `yarn ts` would have blocked the merge.
3. Fixes a Rolldown name-collision regression where the source-level `init_vault()` references would have resolved to `ReferenceError: init_vault is not defined` at runtime (Rolldown auto-renames the plugin-generated factory to `init_vault$1`). The follow-up adds a post-bundle alias.

Thanks for the original analysis and the lazy-accessor fix — that was the right diagnosis and the right mechanism, and it carried straight through into the merged PR.

verseon0980:
Thanks @WiktorStarczewski for the merge and the follow up fixes.

Glad the approach worked well appreciate you covering the remaining cases.

PR #227, opened 2026-05-01
feat: hot key, cold key and guardian key

0xnullifier:


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Features
 
+* [FEATURE][all] 3-key Guardian accounts. Guardian creation now provisions a hot ECDSA key (random, lives outside the WASM keystore behind a new `secure-hot-key` facade — extension/desktop use a JS fallback today; mobile native plugin lands in Phase 4) plus an HD-derived cold ECDSA key (kept in the SDK keystore for cold-routed flows) alongside the existing guardian co-signer. Threshold stays at 1 (hot OR cold + guardian); cold-only routing for `update_signers` / `update_guardian` / `update_procedure_threshold` is enforced client-side per the Phase 0 SDK reading. Storage adds `accColdSecretKeyStrgKey` and `WalletAccount.{hotPublicKey,coldPublicKey}` so role-aware `signWord` (Phase 3) can dispatch hot vs cold by storage entity. Hard cutover from the 1-of-1 Falcon scheme — pre-cutover Guardian accounts are unreachable from this build.
```
WiktorStarczewski: The claim *"cold-only routing for `update_signers` / `update_guardian` / `update_procedure_threshold` is enforced client-side"* is aspirational in this PR — the role-aware `signWord` that does the actual enforcement is deferred to Phase 3. Until then, any signer (hot or cold) can sign these proposals. Practical risk is low (no UI exposes those operations yet), but the changelog reads like the security property holds today; recommend rewording to *"will be enforced client-side once Phase 3's role-aware signWord lands"* to avoid setting a false expectation.

src/lib/miden/back/vault.ts
```diff
@@ -38,6 +39,7 @@ enum StorageEntity {
   MigrationLevel = 'migration',
   Mnemonic = 'mnemonic',
   AccAuthSecretKey = 'accauthsecretkey',
+  AccColdSecretKey = 'accouldsecretkey',
```
WiktorStarczewski, line 109: Typo: `'accouldsecretkey'` — should be `'accoldsecretkey'`. Not user-visible, but this becomes the on-disk storage prefix and gets baked into every cold-key record forever once accounts start writing. Easier to rename now (no live data yet) than after launch.

src/lib/miden/back/vault.ts
```diff
@@ -71,6 +76,27 @@ const insertKeyCallbackWrapper = (passKey: CryptoKey) => {
...
+ *
+ * Storage keys use the unprefixed signer-commitment hex (matching the
+ * WalletSigner convention everywhere else in the codebase).
+ */
+async function persistGuardianKeys(vaultKey: CryptoKey, keys: CreatedGuardianKeys) {
+  await encryptAndSaveMany(
+    [
+      [accAuthPubKeyStrgKey(keys.hotPublicKey), keys.hotPublicKey],
+      [accAuthSecretKeyStrgKey(keys.hotPublicKey), keys.hotCiphertext],
+      [accColdSecretKeyStrgKey(keys.coldPublicKey), keys.coldSecretKeyHex]
```
WiktorStarczewski, line 162: Asymmetric with the hot path: hot writes both `accAuthPubKeyStrgKey(hotPublicKey)` and the secret-key entry, but cold only writes `accColdSecretKeyStrgKey(coldPublicKey)` — no `accAuthPubKeyStrgKey(coldPublicKey)`. If anything in the codebase enumerates "all known auth pubkeys" via the `accAuthPubKey` namespace, cold is invisible to it. Probably intentional (cold is for rotation, not daily signing) but worth a one-line comment so the next reader doesn't think it's a bug.

src/lib/miden/sdk/miden-client-interface.ts
```diff
@@ -104,9 +110,11 @@ export class MidenClientInterface {
...
   async createMidenWallet(walletType: WalletType, seed?: Uint8Array): Promise<string> {
     if (walletType === WalletType.Guardian) {
-      const { createGuardianAccount } = await import('../guardian/account');
-      const account = await createGuardianAccount(this.client, seed);
-      return getBech32AddressFromAccountId(account.id());
+      // Guardian creation needs to surface the hot/cold keys produced inside
+      // createGuardianAccount so the vault can persist them. Callers must use
+      // createGuardianMidenWallet directly — this branch protects against the
+      // older string-returning API being relied on.
+      throw new Error('Guardian wallets must be created via createGuardianMidenWallet');
```
WiktorStarczewski: Runtime throw rather than type-fence. The signature still accepts `WalletType.Guardian`, so accidental calls only blow up at runtime. Better:

```ts
async createMidenWallet(
  walletType: Exclude<WalletType, WalletType.Guardian>,
  seed?: Uint8Array,
): Promise<string>
```

so `tsc` catches the misuse. Same applies to `importAccountBySeed` below.

src/lib/miden/sdk/miden-client-interface.ts
```diff
@@ -131,46 +150,68 @@ export class MidenClientInterface {
   async importAccountBySeed(
     walletType: WalletType,
     seed: Uint8Array,
-    signWordFn: SignWordFunction,
-    getPublicKeyForCommitment: (commitment: string) => Promise<string>
+    _signWordFn: SignWordFunction,
```
WiktorStarczewski: `_signWordFn` / `_getPublicKeyForCommitment` are now unused (Guardian path throws, non-Guardian path doesn't reference them). The `_` prefix says "unused, ignore" but they're still in the call surface and callers still pass them. Either drop from the signature, or keep them un-prefixed with a brief comment that they're retained for caller-compatibility during migration. The current shape is the worst of both worlds.

src/lib/miden/sdk/miden-client-interface.ts
```diff
@@ -131,46 +150,68 @@ export class MidenClientInterface {
...
-          this.client
-        );
-        return getBech32AddressFromAccountId(account.id());
-      } catch (error) {
-        console.log(error);
-        throw new Error('Failed to import Guardian account from seed');
-      }
+      // Guardian import surfaces the hot/cold keys in addition to the account
+      // ID. Use importGuardianAccountBySeed.
+      throw new Error('Guardian wallets must be imported via importGuardianAccountBySeed');
```
WiktorStarczewski: Same recommendation as `createMidenWallet` above — type-fence rather than runtime-throw, so the compiler catches accidental Guardian routing through this entry point.

src/lib/miden/sdk/miden-client-interface.ts
```diff
@@ -131,46 +150,68 @@ export class MidenClientInterface {
...
+   * Caveat (3-key migration): the cold key is fully recoverable from the
+   * mnemonic, but the hot key is random and cannot be reconstructed. This
+   * method generates a fresh hot key, which means the locally-derived
+   * account ID won't match the one the user originally created — recovery
+   * after import requires a `replace_signer` proposal signed by cold +
+   * guardian to re-bind the new hot pubkey on-chain. Phase 8 of the
+   * migration plan covers that flow; until it lands, importing a Guardian
+   * account leaves the user with an empty fresh account under their seed.
+   */
+  async importGuardianAccountBySeed(
```
WiktorStarczewski: **The high-priority concern.** The docstring above is unusually honest about the recovery hole — until Phase 8 lands, this method generates a fresh hot key, derives a different `accountId`, and returns it as if everything worked. The user re-imports their mnemonic expecting their balances back; instead they get an empty new account.

Returning success here is a footgun. Until the Phase 8 `replace_signer` flow ships, recommend either:
- `throw new Error('Guardian account import not yet supported in this build — please re-create the account')`, or
- return success but have the calling vault path surface a UI banner explaining the user has a fresh account and needs to wait for the recovery feature.

Either way the current happy-path return is misleading.

src/lib/miden/guardian/account.ts
```diff
@@ -40,71 +60,98 @@ export async function getSignerDetailsFromAccount(
...
+    const coldSk = AuthSecretKey.ecdsaWithRNG(coldSeed);
+    const coldPublicKeyObj = coldSk.publicKey();
+    const coldCommitmentHex = coldPublicKeyObj.toCommitment().toHex();
+    const coldPublicKey = Buffer.from(coldPublicKeyObj.serialize().slice(1)).toString('hex');
+    const coldSecretKeyHex = Buffer.from(coldSk.serialize()).toString('hex');
+
+    // Hot key — random, held outside the SDK keystore. On extension/desktop
+    // this is the JS fallback (serialized AuthSecretKey hex); on mobile this
+    // will be ECIES-wrapped under SE/StrongBox once Phase 4 lands.
+    const hot = await secureHotKey.generateHotKey();
```
WiktorStarczewski, line 140: This call is unconditional — on iOS/Android `secureHotKey` resolves to `nativePlugin` which throws `NOT_IMPLEMENTED`. So tapping "create Guardian account" on mobile raises a stack trace today. Either gate the Guardian creation UI behind `!isMobile()` until Phase 4, or have `nativePlugin.ts` defer to `jsFallback` as a soft-degradation with a TODO marker. Current behavior is silent regression for any mobile user who finds the Guardian flow.

src/lib/miden/guardian/account.ts
```diff
@@ -40,71 +60,98 @@ export async function getSignerDetailsFromAccount(
...
+
+    // Signer order is [hot, cold] by convention — the migration plan diagrams
+    // and downstream role-routing code assume this layout.
     const multisig = await client.create(
       {
         threshold: 1,
-        signerCommitments: [signerCommitment.toHex()],
-        guardianCommitment: commitment,
-        guardianPublicKey: pubkey,
+        signerCommitments: [hot.commitmentHex, coldCommitmentHex],
```
WiktorStarczewski, line 154: Comment above says *"order is load-bearing for downstream role routing"* — but only the test asserts it. A future contributor sorting the array (for deterministic hashing, dedup, etc.) would silently break role routing. Recommend codifying the invariant:

```ts
const SIGNER_ROLES = ['hot', 'cold'] as const;
// ...
signerCommitments: [hot.commitmentHex, coldCommitmentHex] satisfies readonly string[]
```

or a runtime assertion at signer-key persistence time that maps commitment → role and traps mismatches.

src/lib/miden/guardian/signer.ts
```diff
@@ -6,12 +6,13 @@ export type SignWordFunction = (publicKey: string, wordHex: string) => Promise<s
 export class WalletSigner implements Signer {
   readonly commitment: string;
   readonly publicKey: string;
-  readonly scheme: SignatureScheme = 'falcon';
+  readonly scheme: SignatureScheme;
   private signWordFn: (wordHex: string) => Promise<string>;
 
-  constructor(publicKey: string, commitment: string, signWordFn: SignWordFunction) {
+  constructor(publicKey: string, commitment: string, signWordFn: SignWordFunction, scheme: SignatureScheme = 'ecdsa') {
```
WiktorStarczewski, line 16: Default flipped from hardcoded `'falcon'` to parameter default `'ecdsa'`. Intentional in the cutover, but every existing 3-arg `new WalletSigner(pk, commit, signFn)` call site now silently becomes ECDSA. Worth grepping the base branch (`utk-psm-integration`) for all WalletSigner constructions to confirm the cutover is what's wanted at each one — silent default flips are exactly where unintended behavior changes hide.

src/lib/secure-hot-key/jsFallback.ts
```diff
@@ -0,0 +1,46 @@
...
+  const rawSeed = crypto.getRandomValues(new Uint8Array(32));
+  const sk = AuthSecretKey.ecdsaWithRNG(rawSeed);
+  const ciphertext = Buffer.from(sk.serialize()).toString('hex');
+  const publicKey = sk.publicKey();
+  const publicKeyHex = Buffer.from(publicKey.serialize().slice(1)).toString('hex');
+  const commitmentHex = publicKey.toCommitment().toHex();
+  return { ciphertext, publicKeyHex, commitmentHex };
+}
+
+export async function signHotDigest(ciphertext: string, wordHex: string): Promise<string> {
```
WiktorStarczewski, line 35: `signHotDigest` and `deleteHotKey` have zero test coverage in this PR. They're not invoked yet (Phase 3 wires them up), but the `ciphertext` format is the on-disk blob shape — high-stakes to keep stable across phases. A round-trip test like:

```ts
const hot = await generateHotKey();
const sig = await signHotDigest(hot.ciphertext, '0x' + '00'.repeat(32));
expect(sig).toMatch(/^0x[0-9a-f]+$/);
```

would catch a future refactor that changes the ciphertext encoding and silently breaks loading of existing hot keys.

src/lib/secure-hot-key/nativePlugin.ts
```diff
@@ -0,0 +1,22 @@
...
+ * the gap is visible during testing rather than silently degrading.
+ */
+
+import type { GeneratedHotKey } from './jsFallback';
+
+const NOT_IMPLEMENTED =
+  'secure-hot-key native plugin not yet implemented — Phase 4 of the 3-key migration adds the iOS/Android bridges';
+
+export async function generateHotKey(): Promise<GeneratedHotKey> {
+  throw new Error(NOT_IMPLEMENTED);
```
WiktorStarczewski: Two suggestions for this whole file:
1. Add a tiny test that asserts each export throws with the expected message — so the "not implemented" guard isn't accidentally swallowed by a try/catch upstream.
2. Confirm `lib/platform.isMobile()` resolves to `false` in the test environment so the Guardian-creation tests don't accidentally hit this path. (And see the `account.ts:104` comment — gating the Guardian UI behind `!isMobile()` until Phase 4 would close the user-visible regression.)

src/lib/shared/types.ts
```diff
@@ -274,6 +274,11 @@ export interface WalletAccount {
   isPublic: boolean;
   type: WalletType;
   hdIndex: number;
+  // Set on Guardian accounts created with the 3-key model (hot + cold + guardian).
+  // Absent on non-Guardian accounts and on legacy single-signer Guardian records
+  // produced before the migration; consumers should treat absence as "not 3-key".
+  hotPublicKey?: string;
```
WiktorStarczewski, line 346: Optional fields are non-breaking for non-Guardian accounts (good) but they're also non-breaking for **legacy Guardian accounts created on the pre-cutover Falcon build** — those records decode fine, then fail at sign time because the on-chain account uses Falcon and the wallet now produces ECDSA sigs.

Recommend a load-time check: `type === Guardian && hotPublicKey == null` ⇒ flag the account as "requires re-creation" and surface a banner. Catching the breakage at vault load is much better UX than the user discovering it when their next send fails.

WiktorStarczewski (approved):
Approving — direction is right, the `secure-hot-key` facade is a clean abstraction and the test coverage on the happy path is solid.

Leaving inline recommendations rather than blockers, but flagging three I'd like addressed before this rolls to users:

1. **`importGuardianAccountBySeed` returns success while producing a broken-recovery account** until Phase 8 lands. The docstring acknowledges it; the function surface doesn't. Throwing — or surfacing a clear UI banner — is safer than letting the user think their import worked.
2. **No legacy-Guardian detection.** Pre-cutover Guardian records decode fine but fail at sign time. Detecting `type === Guardian && hotPublicKey == null` at vault load and surfacing a re-creation prompt would catch the failure earlier than the next send.
3. **Mobile crashes on Guardian creation** because `nativePlugin.generateHotKey()` throws. Gate the Guardian flow behind `!isMobile()` until Phase 4 lands, or have `nativePlugin` defer to `jsFallback` as a soft-degradation.

The rest are mechanical cleanups (typo, type-fencing, missing tests for `signHotDigest`/`nativePlugin`, signer-commitment ordering invariant). Empty PR body is also worth filling in given this is a phased migration with a hard cutover.

Issue #134, opened 2026-02-24
Wallet Restoration via Seed Phrase Succeeds Using Repeated Single Word Input


Ivanlomoljo26:
Restoring a wallet through a seed phrase went through even by just using only one repeated word as an input.
See the attached video

https://github.com/user-attachments/assets/8aea06eb-5b18-4f44-a492-7dc630178da9

Ivanlomoljo26:
Cc @WiktorStarczewski @0xnullifier 

## Summary

The onboarding import-from-seed flow validates only that each word appears in the BIP-39 English wordlist — it does **not** verify the BIP-39 checksum encoded in the last word of a real mnemonic. Because any 12 wordlist words clear this check, a user typing a repeated word (e.g. `apple × 12`) passes the "Continue" gate, and downstream key derivation (`bip39.mnemonicToSeedSync` → `derivePath("m/44'/0'/…")`) silently turns the bogus phrase into a deterministic PBKDF2 seed and fresh account. The user believes they're restoring their existing wallet; they're actually creating a brand-new one whose address has no relation to the original. The repo already contains the correct `validateMnemonic` pattern in a sibling component, but that component is dead code — the live flow regressed by dropping the check when the onboarding UI was rewritten (commit `4159cfeb`, 2025-02-03).

## Symptoms

Reproduced on `main` (commit `284e99e`, wallet v1.14.2).

User-visible: any 12 words drawn from the BIP-39 English wordlist pass validation, the Continue button activates, onboarding completes normally, and the user lands on the home screen of a **new** wallet whose address does not match the one they tried to restore. There is no error distinguishing "invalid seed phrase" from "successful import" — no checksum check ever ran.

Persistence scope: every Import-from-seed attempt; reproducible on fresh install across Chrome / iOS / Android. The same gap also affects the ForgotPassword flow, which routes through the same submit action.

## Reproduction steps

1. Fresh install of the wallet (extension or mobile). Launch, tap **Import wallet → Seed phrase**.
2. In the 12-word grid, type `apple` into every cell (or paste `apple apple apple apple apple apple apple apple apple apple apple apple`). Any other wordlist word works (`banana`, `zebra`, etc.).
3. The **Continue** button becomes enabled.
4. Tap **Continue**, set a password, tap through the confirmation screen.

[...]

PR #230, opened 2026-05-05
feat(prove): multi-threaded local proving + speculative pre-prove

WiktorStarczewski:
Web SDK PR: #134

## Summary

- Replaces the SW's single-threaded prove path (~25-40s) with a `chrome.offscreen` document hosting wasm-bindgen-rayon over `navigator.hardwareConcurrency` threads (~5-10s).
- Adds speculative pre-prove: as soon as the SendDetails form is valid (debounced 500ms), the SW kicks off execute + offscreen prove in the background. By Confirm-time the proof is usually already cached and submit + apply runs in ~250ms.
- If the user clicks Confirm while the prove is still mid-flight, the wallet awaits the matching speculation rather than starting a duplicate. If the user changes form params mid-prove, the offscreen doc is terminated to abort the now-stale prove (cores freed for the new params); a non-speculative-prove counter blocks the abort whenever a real send / consume / new-tx prove is in flight, so the user's actual transaction is never killed.
- Drops the per-send "Delegate proving" toggle on the Send screen — local vs delegated is now driven exclusively by Settings → General → Local proving.
- Fixes a pre-existing latent bug in `TransactionProgressModal`: the post-grace pathname auto-dismiss closed the modal during slow proves, so the user landed on Home with no Tx Completed / View on Midenscan screen. Now gated on terminal-state signals (`lastCompletedTxHash`, `transactionComplete`, `hasErrors`).
- Off by default; built-time `MIDEN_USE_OFFSCREEN_PROVING=true` and `MIDEN_USE_SPECULATIVE_PROVING=true` enable it (defaulted on for desktop Chrome via `build:chrome`). Mobile pins both flags to `false` — no equivalent `chrome.offscreen` API on mobile platforms.

Depends on https://github.com/0xMiden/web-sdk/pull/134 for the SDK-side mt-wasm build + `_getInnerWebClient()` escape hatch.

## Test plan

- [ ] `build:chrome` produces a working extension; pack and install in Chrome.
- [ ] Onboard a fresh wallet on testnet, fund the faucet account.
- [ ] Open a Send flow, fill in recipient / amount; observe `[speculation] pre-proved tx in <ms>` in the SW console while still on Review.
- [ ] Confirm the send; verify the modal flips to Tx Completed and the View on Midenscan link works.
- [ ] On Send: change the amount mid-speculation; observe the prior speculation aborts (offscreen doc closes, log shows the abort) and a new prove starts.
- [ ] Confirm a send while the speculation is mid-flight (under load on a slow machine); observe `[mt-offscreen-prove] awaited in-flight speculation <ms> hit=true`.
- [ ] Toggle Settings → General → Local proving off; verify the Send flow uses the remote prover (no offscreen activity, no speculation).
- [ ] Verify a real send is NEVER aborted by a concurrent speculation (manual race: spam form changes during a slow-machine prove).
- [ ] Mobile build (`build:mobile`) still uses remote proving (no offscreen doc, no speculation).


0xnullifier:
Does this do anything on mobile or is it for extension only ?

src/screens/send-flow/ReviewTransaction.tsx
```diff
@@ -79,12 +77,9 @@ export const ReviewTransaction: React.FC<ReviewTransactionProps> = ({
           {/* Options Card */}
           <div className="mt-3">
             <DetailCard>
-              <DetailRow label={t('privatePayment')}>
+              <DetailRow label={t('privatePayment')} isLast={!hasRecall}>
                 <ToggleBadge enabled={sharePrivately} />
               </DetailRow>
-              <DetailRow label={t('delegateProving')} isLast={!hasRecall}>
-                <ToggleBadge enabled={delegateTransaction} />
-              </DetailRow>
```
0xnullifier, line 87: Why this change ?

src/screens/send-flow/SendManager.tsx
```diff
@@ -169,16 +169,96 @@ export const SendManager: React.FC<SendManagerProps> = ({ preselectedTokenId })
...
+  // backs out of the send flow, or the tab closes). Drops any cached
+  // completed entry and marks any active as stale so we don't carry
+  // speculative state into a future send.
+  useEffect(() => {
+    if (process.env.MIDEN_USE_SPECULATIVE_PROVING !== 'true') return;
+    if (!isExtension()) return;
+    return () => {
+      requestSpeculateInvalidate();
+    };
+  }, []);
```
0xnullifier, line 261: What if the user dropped the review screen, changed the params and went to review again. This would trigger another speculate() and there is also an older speculate which is not invalidated as closing the review we are still under the SendManger.

So we wasted whatever time and memory that would have been saved other wise if invalidated as soon as review closes

0xnullifier (changes requested):
I think we should invalidate as soon as the review transaction is closed cause if the user changes the send details then new speculated will be triggered and the older one which is not discarded is wasted memory

src/screens/send-flow/SendManager.tsx
WiktorStarczewski: There is only ever one speculate going on. New speculates rewrite the old ones (which is much easier than dropping existing ones in flight).

src/screens/send-flow/ReviewTransaction.tsx
WiktorStarczewski: Yeah thats the controversial one. So the rationale is that I think nobody will want to change local/delegated on a per-tx level, it pollutes the send screen (needless complexity), and makes the speculative proving harder in the sense that it is not known whether the send is meant to be local proven or delegated. So I think its a win-win(-win), but lets discuss that in the wallet sync, thats a product change ofc.

0xnullifier: yeah I think with the new design's maybe we can remove it ?

src/screens/send-flow/SendManager.tsx
0xnullifier: but the memory/time is still wasted ? I mean if the review screen closes the speculate still happens i.e the proving still happpens which is not required anymore

Or am I missing something

0xnullifier (approved):
LGTM

PR #324, opened 2026-07-08
fix: onboarding flows for extension and settings reveal for passcode based vaults

0xnullifier:


WiktorStarczewski:
Pushed `b16b316f7` addressing the review findings on this PR:

**🐞 Major — passcode auto-submit could fire the same code twice**
`PasscodeEntry`'s auto-submit effect depends on `onSubmit`, `isSubmitting`, and `disabled`. On a failed unlock the parent bumps its attempt counter and hands down a fresh inline `onSubmit`, re-running the effect while the six digits are still present — firing `onSubmit(code)` again for the same passcode. Added a `submittedCodeRef` that records the code at fire time and resets when the code drops below six digits, guaranteeing exactly one submit per distinct completed code.

**🐞 Minor — `EncryptedWalletFileWalletPassword` in-flight state never engaged**
This form submits directly rather than through react-hook-form's `handleSubmit`, so `formState.isSubmitting` was always `false` — the re-entrancy guard, the button spinner, and the passcode auto-submit's `isSubmitting` check were all dead. Replaced it with a local `useState` (mirrors `Unlock.tsx`): set on submit, reset only on error since success navigates away.

**🐞 Minor — reloading on `#create-password` dead-ended onboarding**
Onboarding state (`onboardingType` + the generated/imported seed) is in memory only. A reload on the create-password screen re-ran the hash handler with `onboardingType === null`, walking into `CreatePassword` → `Confirmation` with no seed. Now it restarts from Welcome when `onboardingType` is null, and `onboardingType` is in the effect deps.

**🌐 Localization gap + nits**
The three new keys (`revealSecretPasscodeInputDescription`, `enterYourPassword`, `incorrectPassword`) existed only in `en.json`, so all other locales fell back to English. Added them to the 12 remaining registered locale resources, converted the `zh_TW` source strings to Traditional, and fixed the `en_GB` double-space typo.

Verified: `eslint --max-warnings 0` clean on all changed files; `tsc` clean for the changed files (the two `Cannot find module` errors are unrelated uninstalled deps in my local worktree); all 13 registered locale resources confirmed to contain the 3 keys.

PR #337, opened 2026-07-19
fix: prevent onboarding flash during MV3 service-worker cold-start

hduoc2003:
## Problem

On a browser cold-start (Chrome restarted, the MV3 service worker not yet alive), a
returning user with an existing vault is briefly shown the **onboarding / restore-from-seed**
screen instead of the **Unlock** screen. On slower machines the window is long enough to
interact with, so users think their wallet was lost ("it makes me restore from seed", "the
wallet is no longer available"). The vault is never actually touched - reopening the popup
once the worker is warm shows the password screen and logs in normally.

## Root cause

`WalletStatus.Idle` is overloaded with two different meanings:

- the frontend store's **initial (un-hydrated)** status - "we haven't heard from the backend yet"
  (`src/lib/store/index.ts`), and
- the backend's **"no wallet exists"** status - `vaultExist ? Locked : Idle`
  (`src/lib/miden/back/store.ts`), which the router renders as onboarding.

`toFront()` drops the backend's `inited` flag, so the frontend cannot tell the two apart:
both are `status === Idle`. `PageRouter` routes `!ready` to `<Welcome>`, so during the
service-worker cold-start window (status still the initial `Idle`, backend not yet responded)
it renders onboarding instead of `<Unlock>`.

The store already tracks an `isInitialized` flag (set to `true` on the first backend state
response) but it was never wired into routing.

## Fix

Add a small pure `resolveRootView()` (`src/app/root-view.ts`) that gates the top-level view on
a `hydrated` flag (= `isInitialized`):

```
locked          -> unlock
!hydrated       -> loading   (backend not heard from yet; show the spinner, not onboarding)
!ready (Idle)   -> welcome   (hydrated + no wallet => genuine new user)
default         -> app
```

`hydrated` is exposed through `useMidenContext` and consumed by `PageRouter` for the `*` and
`/` routes. The un-hydrated window now renders the existing loading spinner
(`RootSuspenseFallback`) and resolves to `<Unlock>` (returning user) or `<Welcome>` (new user)
once the backend responds.

## Testing

- Unit test `src/app/root-view.test.ts` covers the cold-start case and the other transitions.
- `tsc` + the unit suite pass; the Chrome extension builds.
- Manual: with the extension built and loaded, killing the service worker and opening a fresh
  page shows the spinner during the un-hydrated window (never onboarding) and then resolves to
  Unlock for an existing vault / Welcome for a fresh install. The normal unlock -> home flow is
  unchanged.

## Notes

- No behavior change for a genuinely new user: they still reach onboarding, just after
  hydration rather than during it.
- Non-breaking; touches only the top-level routing gate.


WiktorStarczewski:
Re-triggering CI: the fork-checkout fix (#345) is now on `main`. This PR doesn't touch `pr.yml`, so its merge ref now picks up the fixed workflow. Closing + reopening to force a fresh `pull_request` run against the updated base (a plain "Re-run" would replay the stale pre-fix merge ref). The new run will need workflow approval again.

WiktorStarczewski (approved):
Thank you for the contribution. LGTM.

PR #349, opened 2026-07-21
feat(guardian): expose guardian info to dApps + reconcile out-of-band guardian switches

### LINKED ISSUE #347: Expose guardian info to dApps + resolve out-of-band guardian switches
## Summary

Expose the connected account's **guardian info** (endpoint, provider, sync status) to dApps via the wallet adapter, and add a wallet-side mechanism to keep each device's guardian metadata correct after an **out-of-band guardian switch**. The two are intertwined — the exposed endpoint is only trustworthy if the wallet keeps its own copy correct — so they're designed and built together.

Motivation: a dApp developer asked for a supported way to read the connected account's guardian endpoint/type. Today `useWallet()` / `useMidenFiWallet()` expose only `address` and `publicKey`, and the dApp bridge carries no guardian data. Separately, if a user switches guardian on another device (e.g. mobile), the extension silently retains a stale endpoint with no way to notice or recover.

## Scope

**Part A — Wallet-side out-of-band detection & resolution**
- On the existing 3s guardian-sync loop, detect a switch by comparing the account's **on-chain guardian commitment** against a new local baseline (`WalletAccount.guardianOperatorCommitment`).
- Resolve for the three built-in operators via a **zero-leak local match**: fetch each operator's public, unauthenticated `GET /pubkey` and compare locally — no account id / signer key / auth is sent to any operator. Auto-heal silently on match.
- For **custom URLs** (unresolvable by design), prompt the user for the URL and **verify it against the on-chain commitment** before persisting.
- Track per-account `guardianSyncStatus` and gate guardian-dependent ops while unresolved.

**Part B — dApp exposure**
- New live request method `requestGuardianInfo()` on **both** `useWallet()` and `useMidenFiWallet()`, returning:
  ```ts
  interface GuardianInfo {
    isGuardianAccount: boolean;
    guardianEndpoint: string | null;               // null: non-guardian / mainnet / unresolved
    guardianProvider: 'open-zeppelin' | 'gateway' | 'lambda-class' | 'custom' | null;
    guardianSyncStatus: 'in-sync' | 'out-of-sync' | null;
  }
  ```
- Free once connected (like `address`/`publicKey`); connected account only; non-guardian accounts return the safe null/false shape.

## Key findings that shape the design

- The guardian endpoint URL is **not derivable** from any account-bound data (on-chain storage, account model, node RPC, or SDK) — verified exhaustively. It lives only in wallet-local storage.
- The on-chain guardian commitment **is** the operator's operator-wide `GET /pubkey`, which is why built-in operators resolve by a leak-free local match. Custom URLs are only recoverable from local storage or a user prompt.

## Non-goals (future, separate work)

- Cross-device guardian-metadata sync.
- An on-chain operator-id/URI commitment (protocol/component change) that would make even custom URLs auto-resolvable and survive operator key rotation.
- Any change to the guardian security model. **No web-sdk change required.**

## Cross-repo

- Adapter plumbing lands in [`0xMiden/wallet-adapter`](https://github.com/0xMiden/wallet-adapter) (new `GuardianInfo` type + `requestGuardianInfo` on base props and both React contexts).
- Detection/resolution + the dApp backend handler land in this repo.

## Status

Design spec written and committed on branch `wiktor/guardian-info-exposure` (`docs/superpowers/specs/2026-07-20-guardian-info-exposure-and-resolution-design.md`); implementation plan in progress.

WiktorStarczewski:
## What this does

Closes #347 

Two intertwined capabilities for Guardian accounts:

**A — Out-of-band guardian resolution.** When a Guardian account's guardian is switched on another device, this device silently held a stale endpoint with no way to notice. Now, on the existing 3s guardian-sync loop, the wallet compares the account's on-chain guardian key commitment against a locally-persisted baseline (`guardianOperatorCommitment`) and, on drift:
- **Auto-resolves** for the three built-in operators via a **leak-free local match** — it fetches each operator's public, unauthenticated `GET /pubkey` and compares locally; no account id / signer key / auth is sent to any operator.
- Falls back to a **verified user prompt** for custom URLs (a banner that validates the entered URL against the on-chain commitment before persisting).
- Gates guardian-dependent transactions while out of sync (`assertGuardianInSync`), exempting the guardian-switch/recovery path so recovery can't deadlock.

**B — dApp exposure.** A read-only `requestGuardianInfo()` request returning the connected account's `{ isGuardianAccount, guardianEndpoint, guardianProvider, guardianSyncStatus }`. Non-guardian accounts get a null/false shape. It's a direct-return handler (no confirmation prompt — guardian info is non-sensitive), so it works uniformly on extension/mobile/desktop.

## Why the endpoint can't just be derived

The guardian endpoint URL is never on-chain (only the operator's key commitment is, in private storage) and isn't in the node RPC or SDK — verified exhaustively. The on-chain commitment **is** the operator's operator-wide `GET /pubkey`, which is exactly why built-in operators resolve by a zero-disclosure local match, and why custom URLs can only come from local storage or a user prompt.

## Cross-repo

The dApp-facing adapter method lands in a companion PR on [`0xMiden/wallet-adapter`](https://github.com/0xMiden/wallet-adapter) (`requestGuardianInfo` on `useWallet`/`useMidenFiWallet`). The two integrate over the string wire protocol (`GUARDIAN_INFO_REQUEST`), not shared types, so neither blocks the other's CI — but both must land for the end-to-end dApp feature.

## Testing

- Full suite **2479/2479** passing.
- Coverage **95.82 / 95.16 / 95.63 / 95.82** — clears the global 95/95/95/95 gate.
- `yarn lint` + `yarn lint:i18n` clean.
- Built TDD-first, task-by-task, with independent per-task review and two whole-branch reviews. One review-found bug (a partial-write could strand an account in `resolving` with no recovery UI) is fixed with regression tests.

## Open question for review

The standalone `setGuardianOperatorCommitment` / `setGuardianSyncStatus` store actions + `client.ts` exposure are currently unused (resolution writes through the vault directly). Keep them (follows the repo's "adding a wallet action" convention, tested, harmless) or trim them (YAGNI)? Happy to remove if preferred.

Part of #347.


src/lib/miden/back/guardian-drift.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * The WASM account read is lock-guarded; the built-in-operator HTTP probe
+ * runs outside the lock.
+ */
+export async function resolveGuardianDrift(
+  vault: GuardianDriftVault,
+  accountPublicKey: string
+): Promise<{ status: GuardianSyncStatus; changed: boolean }> {
+  const account = await vault.getAccount(accountPublicKey);
+  if (!account) return { status: 'in-sync', changed: false };
```
0xnullifier, line 50: nit: I think we should throw an error here

0xnullifier (approved):
LGTM, just a nit

PR #351, opened 2026-07-21
fix(ios): biometric-only wallets can't lock out (App Store 2.1 rejection)

WiktorStarczewski:
## Problem — App Store rejection (Guideline 2.1(a), build 1.15.6)

On an iPad Air (M3) the reviewer reported the app *"constantly displayed the Touch ID required message."* Root cause (confirmed in code): a **biometric-only ("passwordless") wallet has no non-biometric way in, and the escape hatch is broken.**

- Choosing the primary "Use Face ID or Biometric" onboarding button creates a hardware-only vault with **no app passcode** (`Welcome.tsx:193`, `vault.ts` `setupHardwareProtector`).
- The most likely trigger is an **orphaned Secure Enclave key**: onboarding persists the SE key **before** encrypting the vault blob, so a cancelled setup leaves a key with no blob → unlock throws instantly → the "Biometric Unlock Required" screen, whose "Try Again" repeats and whose **"Reset Wallet" didn't even delete the SE key** (iOS reset only cleared Preferences) → truly stuck.

## Fix (defense-in-depth, preserves the passwordless design)

- **`LocalBiometricPlugin.swift`** — the vault Secure Enclave key gains `.userPresence` (matching the hot key), and the unlock decrypt now gates on `.deviceOwnerAuthentication` (**biometric OR device passcode**), reusing one authenticated context so there's a single prompt. A passwordless wallet on a device with no enrolled biometrics can now unlock with the **device passcode**.
- **`vault.ts`** — `setupHardwareProtector` deletes the just-generated SE key if setup fails (e.g. the user cancels the auth prompt), so no orphaned key is left behind.
- **`reset.ts`** — iOS reset now deletes the Secure Enclave key (was desktop-only), so a stuck wallet can actually recover.
- **`Unlock.tsx`** — relabel the recovery button to "Unlock" (the passcode path is now real).

## Verification

- **TypeScript fully verified**: `tsc` clean; existing `reset`/`vault` suites green; new tests added (orphaned-key cleanup, reset-deletes-SE-key, best-effort tolerance); **coverage clears the 95% gate** (95.80 / 95.06 / 95.44 / 95.80).
- **Swift reviewed but NOT device-tested** — the iOS *simulator force-disables hardware security*, so the biometric-only path cannot be exercised in the sim, and I had no real device. **Before merging/resubmitting, build in Xcode and device-test:** an iPhone/iPad with **biometrics NOT enrolled** + a **device passcode set** → fresh install → onboard with "Use Face ID or Biometric" → lock → confirm unlock via **device passcode**; also cancel setup mid-onboarding + relaunch and confirm no dead-end and that "Reset Wallet" recovers.

## Notes

- PR targets `wiktor/appstore-1.15.6` (the resubmission branch). The same fix should also be ported to `main` — the bug is in core code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+* [FIX][ios] **Biometric-only wallets can no longer get permanently locked out (App Store 2.1).** A hardware-only (passwordless) vault is now unlockable with the device passcode when biometrics can't be satisfied — the Secure Enclave key gains `.userPresence` and the unlock decrypt gates on `.deviceOwnerAuthentication` (biometric OR passcode). Onboarding no longer leaves an orphaned Secure Enclave key if setup is cancelled (it's deleted on failure), and iOS "Reset Wallet" now deletes the Secure Enclave key (previously desktop-only), so a stuck wallet can actually recover. Fixes the iPad review-lab lockout where the app "constantly displayed the Touch ID required message".
+
 * [FIX][devnet] **Restored the devnet developer (wrench) badge on the icon.** The v0 UI revamp (#248) rebranded the devnet icon to the plain Bread "B" and dropped the wrench badge, so devnet builds again looked identical to production. `public/misc/logo-devnet*.png` (234/128/48/40/32/16) now re-add the Advanced Settings / Developer wrench glyph in a white circle with a thin `#7286A0` ring, overlapping the B's bottom-right, so a devnet build is distinguishable at a glance. Wiring is unchanged — `vite.extension.config.ts` already swaps `logo-white-bg*` → `logo-devnet*` for `MIDEN_NETWORK=devnet` (manifest icons, action icon, and tab favicon).
 
 ## 1.15.6 (2026-07-09)
--- a/ios/App/App/LocalBiometricPlugin.swift
+++ b/ios/App/App/LocalBiometricPlugin.swift
@@ -255,15 +255,21 @@ public class LocalBiometricPlugin: CAPPlugin, CAPBridgedPlugin {
         ]
         SecItemDelete(deleteQuery as CFDictionary)
 
-        // Create access control - only require auth when USING the private key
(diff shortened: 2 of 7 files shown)


WiktorStarczewski:
## Review pass (multi-lens, adversarial) — one critical regression caught and fixed

Ran an adversarial code review (Swift/Secure-Enclave, TS logic/regression, root-cause completeness, and a "try to still get locked out" lens) with per-finding verification, since this can't be device-tested (the iOS simulator force-disables hardware security).

**🔴 Critical (fixed in `2ce8a367f`):** the earlier revision added an explicit `LAContext.evaluatePolicy(.deviceOwnerAuthentication)` gate before the Secure Enclave key op. On a device with **no passcode set**, that returns `passcodeNotSet` → hard-fail — which would **brick pre-fix wallets** whose `.privateKeyUsage`-only keys previously decrypted via the key op alone. It also carried a latent double-prompt risk. Fix: removed the explicit gate; the `.userPresence` ACL on the key already makes the key op itself prompt for **biometric OR device passcode** (single native prompt, passcode fallback) — so existing keys keep their exact behavior and new keys gain the fallback. Also added a `SecKey` type guard before the force-cast (parity with HotKeyPlugin), and set the orphan-cleanup flag before `generateHardwareKey`.

**Verified clean:** `tsc` clean, 30/30 touched-file tests pass, coverage-neutral. Swift can't be compiled here (no synced Pods) — reviewed by reading; braces/logic check out.

**Still requires REAL-DEVICE verification before resubmit** (review cannot settle these): on a physical iPhone/iPad with **biometrics NOT enrolled + a device passcode set** → fresh install → onboard with "Use Face ID or Biometric" → lock → confirm unlock via **device passcode** and that it's a **single** prompt. Also: cancel setup mid-onboarding + relaunch → no dead-end; "Reset Wallet" recovers.

Non-blocking follow-up (not this PR): consider a seed-backup gate before allowing passwordless mode, so a device-passcode change can never strand funds.

PR #358, opened 2026-07-21
fix: route guardian dApp txs correctly (normalize account id vs composite publicKey)

WiktorStarczewski:
## Problem

A dApp transaction from a **Guardian** account fails at execute time with:

```
transaction execution failed: transaction is unauthorized with summary TransactionSummary { … }
```

Reproduced on **miden.name** (name registration → a network note): every attempt from a guardian account failed, while single-sig accounts and in-wallet guardian sends worked. It is **not** attachments, the tx summary, the guardian-server version, or the network-account setup.

## Root cause

The failure is a **guardian routing** miss caused by an account-id **format mismatch**:

- A dApp/adapter supplies the account as its **bare bech32 address**, e.g. `mtst1arqnh74pn2m37sff0w30uvmfdc5068r3`.
- `WalletAccount.publicKey` is a **composite** `<address>_<suffix>`, e.g. `mtst1arqnh74pn2m37sff0w30uvmfdc5068r3_qr7qqq9wr6w`.

`isGuardianAccount` and `getOrCreateMultisigService` (`front/guardian-manager.ts`) matched them with a raw `acc.publicKey === accountId`, which **misses** on the dApp form. `generateTransaction` then routes the tx down the **non-guardian** path (`midenClient.newTransaction`), so **no guardian co-signature is attached** and the on-chain guardian auth (`verify_signatures`) finds zero guardian sigs → `AUTH_UNAUTHORIZED`.

This explains the full symptom set:
- **In-wallet guardian sends work** — they use the wallet's own composite `publicKey` → exact match → routed guardian → co-signed.
- **Single-sig works** — non-guardian routing is correct for it; the id mismatch is irrelevant.
- **dApp guardian txs fail** — bare address → routing miss → no co-sign → unauthorized.

## Fix

- Add a shared **`sameWalletAccountId(a, b)`** (+ `canonicalWalletAccountId`) in `sdk/helpers.ts` that canonicalizes both ids to the SDK account id derived from the address portion (stripping the `_suffix`), with a safe fallback.
- `isGuardianAccount` and `getOrCreateMultisigService` now match via `sameWalletAccountId` instead of raw `===`.
- `getOrCreateMultisigService` loads the SDK account by the **matched account's stored `publicKey`** (the form the working in-wallet path uses), not the possibly-bare dApp id.

The dApp is correct to send the bech32 address; the wallet is the side that stores a composite and must resolve it.

## Verified

- Reproduced end-to-end with an instrumented build: pre-fix the tx was routed non-guardian and failed `AUTH_UNAUTHORIZED`; **post-fix it routes guardian, co-signs, and the miden.name registration succeeds** (`[TransactionProcessor] Loop result: true`).
- `yarn ts` clean, `yarn lint` clean on the changed files, and the two affected suites pass (`guardian-manager.test.ts`, `helpers.test.ts`) — includes a regression test (bare address → routes guardian) and `sameWalletAccountId` unit tests covering the SDK-canonical and fallback paths.

## Notes

- Scope is the two guardian-routing lookups on the dApp critical path. Other `publicKey === id` sites exist (e.g. in `vault.ts`) but are internal flows that consistently use the composite `publicKey`; routing the dApp-facing ones through `sameWalletAccountId` as a follow-up would harden against recurrence on other guardian entry points.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Fixes
 
 * [FIX][mobile] **iOS now renders at the display's native refresh rate (up to 120Hz ProMotion) instead of being capped at 60fps.** Added `CADisableMinimumFrameDurationOnPhone` to `Info.plist`. Since iOS 15, iPhones cap every app — including its WKWebView — to 60fps unless the app opts in to high frame rates, so on ProMotion devices animations and scrolling felt stuck at 60Hz (the same as Low Power Mode). CSS/compositor animations and native scrolling now run at the full display rate.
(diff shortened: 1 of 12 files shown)


0xnullifier (approved):
LGTM

PR #364, opened 2026-07-22
fix: restore encrypted wallet file into the active miden-client store (#253)

### LINKED ISSUE #253: Encrypted-wallet-file restore writes to 'miden-wallet' but active client reads from 'MidenClientDB_<network>' — balances stay 0 (regression from #151)
## Issue Description

Restoring a wallet from an encrypted `.miden-wallet` file on v1.14.3 / v1.14.4 leaves the account balance permanently at zero on testnet, even though the account's assets are visible on Midenscan and the wallet's own transaction history renders correctly. The miden-client never sees the restored state because the restore path writes the imported IndexedDB dump to a different database name than the active `MidenClient` instance reads from.

The export uses `this.client.storeIdentifier()` (the active client's actual DB name, default `MidenClientDB_<network_id>`). The import hardcodes `'miden-wallet'` instead. The two never line up, so the running client keeps reading from its own empty default DB.

## Root cause (1 location)

[`src/screens/onboarding/import-wallet-flow/ImportWalletFile.tsx#L104`](../blob/main/src/screens/onboarding/import-wallet-flow/ImportWalletFile.tsx#L104):

```ts
await importStore(midenClientDbContent, 'miden-wallet');
await importDb(walletDbContent);
```

`useImportStore` from `@miden-sdk/react` resolves this to `sdkImportStore('miden-wallet', midenClientDbContent)`, which creates a fresh `IdxdbStore` named `'miden-wallet'` and dumps the payload into it ([`useImportStore.ts:66`](https://github.com/0xMiden/web-sdk/blob/main/packages/react-sdk/src/hooks/useImportStore.ts#L66), [`import.rs:107-120`](https://github.com/0xMiden/web-sdk/blob/main/crates/web-client/src/import.rs#L107-L120)).

Meanwhile the wallet creates its active `MidenClient` without passing `storeName` ([`miden-client-interface.ts:118-152`](../blob/main/src/lib/miden/sdk/miden-client-interface.ts#L118-L152)). The web SDK falls back to the default at [`web-client/src/lib.rs:256-257`](https://github.com/0xMiden/web-sdk/blob/main/crates/web-client/src/lib.rs#L256-L257):

```rust
let store_name =
    store_name.unwrap_or(format!("{}_{}", BASE_STORE_NAME, endpoint.to_network_id()));
```

with `BASE_STORE_NAME = "MidenClientDB"`. So the active client reads from `MidenClientDB_mtst` (or similar) and never observes the data in `'miden-wallet'`.

This is consistent with the wallet's own `exportDb` / `importDb` methods which both use the dynamic name correctly ([`miden-client-interface.ts:543-551`](../blob/main/src/lib/miden/sdk/miden-client-interface.ts#L543-L551)):

```ts
async exportDb() {
  const storeName = this.client.storeIdentifier();
  return await exportStore(storeName);
}

async importDb(dump: string) {
  const storeName = this.client.storeIdentifier();
  await importStore(storeName, dump);
}
```

## Regression provenance

`git log -S "await importStore(midenClientDbContent, 'miden-wallet');"` points at commit `a22e77e` (PR #151, "Migrate to React SDK and v14 Miden Client", 2026-04-09). That PR replaced the previous `midenClient.importDb(midenClientDbContent)` call — which routed through `storeIdentifier()` correctly — with the hardcoded form. Affects releases v1.14.3 and v1.14.4.

## Reproduction Steps

1. Install wallet v1.14.3 or v1.14.4.
2. Create a testnet wallet, fund it via the testnet faucet, wait for consume to complete.
3. Verify on Midenscan that the account holds assets.
4. Export an encrypted wallet file.
5. Either uninstall + reinstall the extension, OR remove the wallet's databases from DevTools → Application → IndexedDB.
6. Restore from the encrypted file.

Observed: transaction history renders (because `walletDbContent` is correctly restored into the wallet's own `TridentMain` Dexie DB on line 105), but balance shows 0 and console logs `No account header record found for given ID.` on every sync.

Midenscan-side verification of one affected account: <https://testnet.midenscan.com/account/mtst1apmq55qz3dn8uyq3j63jmj2wlga2ntel> (reported by a pioneer on the testnet Discord — they can repro on demand if helpful).

## Suggested direction

Three workable fixes, in increasing order of invasiveness:

1. Revert the call site to `midenClient.importDb(midenClientDbContent)` (routes through `storeIdentifier()`). One-line change but adds back the WASM-client lock dependency, which is what PR #151 was trying to escape.
2. Read the store name from the active client at the call site: `const storeName = (await getMidenClient()).client.storeIdentifier(); await importStore(midenClientDbContent, storeName);`. Preserves the lock-free pattern.
3. Pin `storeName: 'miden-wallet'` in `MidenClientInterface.create` so both ends agree. Cleanest, but requires a one-time migration for existing users whose data is in `MidenClientDB_<network>`.

Happy to send a PR for (2) if that's the preferred direction.

## Related

- #136 "Wallet Recovery via Encrypted File Stuck on 'Your Wallet is Ready!' Screen" — different symptom (stuck on screen rather than zero balance), but same restore flow; possibly same underlying cause if the restored state being unreachable trips a downstream wait.

WiktorStarczewski:
## Summary

Restoring an encrypted wallet file wrote the miden-client IndexedDB dump into a hardcoded store named `miden-wallet`. The running client never reads from that store, so after a successful restore the wallet showed accounts but every balance stayed at 0.

## Root cause

The export path (`MidenClientInterface.exportDb`) serialises the client store using the client's own `storeIdentifier()`, which resolves to `MidenClientDB_<network>`. The restore path in `ImportWalletFile` ignored that and imported the dump into a literal `'miden-wallet'` store instead. The dump landed in a store the active client never opens, leaving the real client DB empty, so account and balance state was invisible and balances read as 0.

## Fix

Resolve the destination store name from the active client the same way the export path does — `(await getMidenClient()).client.storeIdentifier()` — and pass that to `importStore` instead of the hardcoded literal. This guarantees the restore writes into the exact store the live client reads from, regardless of network.

## Testing

- Updated the existing decryption-flow assertion to expect the resolved store name.
- Added a regression test asserting the restore targets the active client's `storeIdentifier()` and never the hardcoded `miden-wallet`.
- `npx jest src/screens/onboarding/import-wallet-flow/ImportWalletFile.test.tsx --no-cache --runInBand` — 29 passed.
- `npx eslint` on the changed files — clean, 0 warnings.

Closes #253


0xnullifier (approved):
LGTM

PR #365, opened 2026-07-22
fix: mark guardian value-moving tx Completed when submit lands but local apply fails (#356)

### LINKED ISSUE #356: Guardian consume can finalize on chain but remain Failed locally after post-submit local apply failure, causing local TST conservation loss
## Summary

  A Guardian-enabled public-note consume finalized on Miden testnet, but the wallet marked the consume transaction as Failed after local apply threw:

  `failed to apply transaction result: account 0x01c334483aeaa7a173c798ec5e8944 is locked because the local state may be out of date with the network`

  Because the Guardian consume path cancelled the transaction instead of reconciling the already-finalized chain transaction, wallet A never locally imported/applied
  the consumed note. This produced an exact local conservation loss of `7 TST`.

  This is a wallet-side Guardian recovery bug. Chain state shows the note was consumed by wallet A.

  ## Run

  - Run: `stress-20260721-062414`
  - Network: testnet
  - Guardian provider: OpenZeppelin, `https://guardian.openzeppelin.com`
  - Wallet version in trace: `1.15.7`
  - Stress loops requested/completed: `300/300`
  - Local conservation result: initial `6000 TST`, final `5993 TST`, delta `-7 TST`

  ## Exact note

  - Note ID: `0x0ca618d0e86955b7a1e2a4a34fe49b1d7d15306ede097245da853dd989292f5d`
  - Nullifier: `0x77dca29357c896da712f68c4bf9c9d2d1384102bcad656d4f8d6b01c7258eb6a`
  - Amount: `7 TST`
  - Flow: B -> A, public note
  - Stress op: `215`

  ## Evidence

  ### 1. Conservation math is exact

  `stress-summary.json` reports:

  - Initial: A `3000`, B `3000`, total `6000`
  - Final: A `3032`, B `2961`, total `5993`
  - Delta: `-7`

  Independent final DB reconciliation found exactly one sent output absent from receiver-side local completion/import accounting, and its amount is exactly `7 TST`.

  ### 2. B created and completed the send

  Final B app DB has completed send transaction:

  - Tx id: `d5438edc-1b41-43a8-809c-835a5d8d01d8`
  - Type: `send`
  - Status: Completed
  - Amount: `700000000`
  - Output note IDs include `0x0ca618d0e86955b7a1e2a4a34fe49b1d7d15306ede097245da853dd989292f5d`
  - Stress metadata: op `215`, B -> A, public

  ### 3. The note exists on chain

  `rpc.Api.GetNotesById` for the candidate note returns:

  - Note type: `NOTE_TYPE_PUBLIC`
  - Sender: B account `OXFjlC+e5CE1ONy6BMtl`
  - Inclusion block: `794741`

  ### 4. The exact nullifier was consumed on chain by wallet A

  `rpc.Api.SyncTransactions` for wallet A account `AcM0SDrqp6Fzx5jsXolE`, block range `795590..795610`, returns:

  - Block: `795599`
  - Chain tx: `0x2a400e679d9a53924deff9df2fbcda0e61562ff86f66727a429e805039682243`
  - Account: wallet A
  - Input nullifier: `0x77dca29357c896da712f68c4bf9c9d2d1384102bcad656d4f8d6b01c7258eb6a`

  This proves the consume finalized on chain under A’s account.

  ### 5. A did not complete or import the consume locally

  Final A app DB has only two consume rows for this note, both Failed:

  - `1da7a1ce-90ed-4c51-84c9-a42f194889fb`
    - Failed because queued tx expired.
  - `9a7bd4a4-23cb-4f69-a3ac-f3a403e45cba`
    - Failed with local apply error:
    - `failed to apply transaction result: account 0x01c334483aeaa7a173c798ec5e8944 is locked because the local state may be out of date with the network`

  Final A local Miden DB has no transaction row matching chain tx `0x2a400e679d9a53924deff9df2fbcda0e61562ff86f66727a429e805039682243`.

  ### 6. Trace lifecycle shows wallet cancellation after Guardian submit/local apply failure

  In `note-trace-A.ndjson`:

  - Candidate note discovered/committed locally.
  - Retry consume tx `9a7bd4a4-23cb-4f69-a3ac-f3a403e45cba` starts.
  - `wallet.guardian.submit` starts for `0x0ca618...292f5d`.
  - `wallet.wasm.call.errored` fires under `guardian_submit`.
  - `generateGuardianTransaction` transitions tx from Generating to Failed.
  - `wallet.tx.recovery.transaction_cancelled` is emitted.
  - No `wallet.guardian.complete` / `consume_committed` event exists for this note.

  ## Current behavior

  - Guardian transactions route through `generateGuardianTransaction`.
  - Structural Guardian operations have special `ApplyTransactionAfterSubmitFailed` recovery.
  - Ordinary Guardian `consume` does not.
  - Non-recovered Guardian errors fall through to `await cancelTransaction(transaction, error, 'generateGuardianTransaction', traceOptions);`
  - The generic non-Guardian loop already has apply-after-submit reconciliation for send/consume via `if (errorCode === 'ApplyTransactionAfterSubmitFailed')`.
  - Guardian consume does not reach that generic handler because the Guardian branch catches and cancels first.
  - `verifyStuckTransactionsFromNode()` only scans `GeneratingTransaction` rows, so once this tx is marked Failed it is excluded from later node-based recovery.

  ## Expected behavior

  If a Guardian consume reaches the node and finalizes on chain, the wallet must not leave the transaction Failed locally.

  The wallet should either:

  - complete the transaction with the chain tx id/result when available, or
  - reconcile via note/nullifier/account sync and mark the local consume as completed/externally consumed, not Failed.

  ## Actual behavior

  The chain consumed the note under wallet A, but the wallet cancelled the local Guardian consume as Failed after local apply threw. The local transaction did not
  reconcile the consumed `7 TST`, so local stress conservation ended at `5993` instead of `6000`.

  ## Impact

  This causes local wallet balance/state drift after Guardian consume transactions: funds are consumed on chain but not reflected locally, producing an apparent local
  token loss.

  ## Suggested fix

  Extend Guardian recovery for ordinary `consume` and likely `send` transactions when submit reached the node but local apply failed.

  Suggested durable behavior:

  - Persist or recover the submitted chain transaction id when possible.
  - On local apply failure after submit, query/sync the account/nullifier/note state before marking Failed.
  - If the input note nullifier is consumed by the wallet account, mark/reconcile the consume as Completed or ConsumedExternal.
  - Do not let a chain-finalized Guardian consume remain terminal Failed.
  - Add regression coverage for Guardian consume where `midenClient.client.transactions.submit()` submits/finalizes but throws during local apply.

  ## Ruled out

  - Not RC-2/private transport loss: the note is public and A discovered/committed it locally.
  - Not sender drop/rejection: the note exists on chain.
  - Not claim-tail accounting: pending tail is a separate `218 TST` set and does not explain the exact `-7`.
  - Not mixed-asset accounting: this row is TST-only.
  - Not merely RPC submit acceptance: A-scoped `SyncTransactions` proves finalized chain consumption of the exact nullifier.

WiktorStarczewski:
## Summary

Fixes a case where a Guardian **consume** (or other value-moving op: send/swap/execute) could be left terminally `Failed` even though its transaction had already landed on chain. The note was spent on chain but the wallet showed the operation as failed, producing an apparent conservation loss that no later sync could recover.

## Root cause

When a Guardian value-moving transaction is submitted, the SDK first submits to the network and then applies the resulting state change locally. If the submit lands on chain but the **local apply** throws (`ApplyTransactionAfterSubmitFailed`), the transaction is live on chain but the local client failed to record it.

Guardian ops are routed through the guardian branch of `generateTransaction`, whose own `catch` handled two cases: structural ops (via `reconcileStructuralApplyFailure`) and the eventually-consistent nonce-overwrite case. A value-moving op that hit `ApplyTransactionAfterSubmitFailed` fell through to the generic cancel path and was marked `Failed`. Because `verifyStuckTransactionsFromNode` only scans in-progress rows, it could never recover a `Failed` row — so the tx stayed terminally failed while the note was already consumed on chain.

The generic `generateTransactionsLoop` path already handled this correctly (mark `Completed`, let the next sync reconcile via `ConsumedExternal`), but Guardian ops never reach that path.

## Fix

In the guardian branch of `generateTransaction`, add an explicit handler that mirrors `generateTransactionsLoop`'s generic `ApplyTransactionAfterSubmitFailed` behavior for value-moving ops (`consume`/`send`/`swap`/`execute`):

- Mark the transaction `Completed` (with `displayMessage` `Claimed` for consume, `Sent` otherwise) instead of cancelling it.
- The next sync reconciles the note state via `ConsumedExternal`.
- Re-marking is wrapped so an already-finalized tx is a no-op rather than an error.

Structural ops are handled earlier in the same catch and never reach this branch on success. The `generateTransactionsLoop` comment is updated to reflect that all Guardian ops (structural and value-moving) are now handled in the guardian branch.

## Testing

- Added two unit tests in `transactions.guardian.test.ts`:
  - Guardian **consume** apply-after-submit failure marks the tx `Completed` (`displayMessage` `Claimed`), not `Failed`/cancelled.
  - Guardian **send** apply-after-submit failure marks the tx `Completed` (`displayMessage` `Sent`).
- Full guardian suite passes: `31 passed, 31 total`.
- `eslint` clean on the changed files (`--max-warnings 0`).

Closes #356


0xnullifier (approved):
LGTM

PR #366, opened 2026-07-22
fix: let side panel shrink below 360px so Brave sidebar does not clip wallet (#336)

### LINKED ISSUE #336: Wallet right edge clipped in Brave — side panel enforces a 360px minimum width that Brave's sidebar doesn't honor
## Starting state
- Wallet installed in Brave with an existing testnet account, opened in the Brave Sidebar (the `chrome.sidePanel` surface, i.e. `sidepanel.html` — not the toolbar popup)
- Brave's own vertical icon rail visible on the same edge as the panel (default Brave setup)

## Steps
1. In Brave, open the Miden Wallet in the sidebar (side panel surface)
2. Observe the Home screen — no interaction needed

## Expected
The wallet layout fits the width Brave gives the panel: full balance visible, all three action buttons (Send / Receive / Faucet) fully visible and clickable, token rows showing their amounts.

## Observed
- The right portion of the wallet UI is cut off flush at the panel edge; Brave's icon rail sits where the missing content should be
- Balance "$0.00" renders as "$0.0" (last glyph clipped)
- The Faucet button is roughly half visible
- Token row fiat values are clipped (e.g. "$0.0" instead of "$0.00")
- No horizontal scrollbar appears, so the hidden content is unreachable
- Brave's sidebar settings offer no way to recover the missing width

## Does NOT reproduce on
- **Chrome** side panel, same build — layout fits with no clipping (key contrast axis: browser)
- Brave **toolbar popup** (`popup.html`) — popup hosts honor the requested 360px, so the fixed-size popup renders fully
- **Full-page view** (`fullpage.html` in a tab) — responsive, no clipping

## Evidence
- Screenshot (attached): Brave sidebar showing the wallet with the right edge clipped under Brave's icon rail — balance, Faucet button, and token amounts all cut
- Source: `sidepanel.html` declares `width: 100%` but also `min-width: 360px` on the root element, and `overflow: hidden` on the body — verified at `main` @ `0294ca29`
- Measured panel viewport width in Brave: [TBD — value of `window.innerWidth` from the panel's DevTools console; expected < 360]

## Notes / hypotheses (speculation, with confidence)
- **Mechanism (high confidence, source-verified):** the side-panel document refuses to shrink below 360px while its body hides all overflow. Any host that provides less than 360px therefore clips the right side with no scrollbar. Chrome's built-in side panel never goes below ~360px, so the floor is never hit there; Brave hosts extension panels inside its Sidebar next to its own icon rail, which eats horizontal space, and Brave's panel resizing is unreliable (upstream: brave/brave-browser#51271, brave/brave-browser#43507) — so in Brave the panel viewport lands below 360px and clipping is guaranteed.
- **Not a duplicate of #249 (closed):** that issue was about the balance *font size* overflowing its own container at large values, on all surfaces. Here the *whole document* is wider than the host viewport — the clipping hits buttons and token rows too, and occurs at $0.00. The #249 auto-fit fix does not and cannot address this.
- **Suggested direction (plain language, long-term):** make the side-panel layout tolerate widths below 360px instead of clamping — let the layout shrink gracefully (the action-button row already flexes; the balance text should scale or truncate), and as a last-resort fallback allow horizontal scrolling rather than silently hiding content. The wallet should not depend on every Chromium fork guaranteeing a minimum panel width.

## Cross-references
- #122 — side panel feature origin
- #249 — closed balance-overflow fix; distinct (see hypotheses)
- brave/brave-browser#51271 — Brave sidebar rail consumes panel width, no option to hide
- brave/brave-browser#43507 — Brave side panels can't be resized reliably

WiktorStarczewski:
## Summary

The side panel could not shrink below 360px wide. In Brave, the browser's side panel renders narrower than ~360px (its icon rail eats into the available width), so the hard `min-width: 360px` on the root `<html>` forced the document wider than the host viewport. Because the side panel is styled with `no-scrollbar`, the overflow produced no horizontal scrollbar — the clipped right edge (balance glyphs, the Faucet button, token fiat values) simply became unreachable.

## Root cause

`public/sidepanel.html` set an inline `min-width: 360px` on the root `<html>` element. When the host viewport is narrower than 360px, that floor forces horizontal overflow, and the `no-scrollbar` styling hides the scrollbar that would otherwise let the user reach the clipped content.

## Fix

Change the root `<html>` inline style from `min-width: 360px` to `min-width: 0`, letting the document shrink to match the host viewport. The existing responsive layout inside the panel already reflows down to narrow widths, so removing the artificial floor lets the content fit within Brave's sub-360px side panel with no clipping.

Added `src/sidepanel.html.test.ts` as a regression guard asserting the root `<html>` no longer clamps its width to 360px.

## Testing

- `npx jest src/sidepanel.html.test.ts --no-cache --runInBand` — 2 passing tests.
- `npx eslint src/sidepanel.html.test.ts --max-warnings 0` — clean.

Note: this is a UI/CSS change that was NOT visually verified in this environment. Manual visual verification (Brave sidebar at sub-360px width for #336) is recommended before and after merge to confirm the wallet content fits without clipping.

Closes #336

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.15.9 (TBD)
+
+### Fixes
+
+* [FIX][extension] **The side panel no longer clips its right edge in Brave's narrow sidebar.** The side-panel document hard-clamped its root to `min-width: 360px` with hidden overflow, so any host giving less than 360px (Brave's sidebar, whose icon rail eats width) clipped the balance, Faucet button, and token amounts with no scrollbar. The floor is removed so the layout shrinks to the host viewport.
+
 ## 1.15.8 (2026-07-21)
 
 ### Features
--- a/public/sidepanel.html
+++ b/public/sidepanel.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="" class="no-scrollbar" style="width: 100%; height: 100%; min-width: 360px; background-color: #FFFFFF;">
+<html lang="" class="no-scrollbar" style="width: 100%; height: 100%; min-width: 0; background-color: #FFFFFF;">
   <head>
     <meta charset="utf-8" />
     <meta
--- a/src/sidepanel.html.test.ts
+++ b/src/sidepanel.html.test.ts
@@ -19,7 +19,7 @@ describe('public/sidepanel.html', () => {
 
   const rootStyle = (() => {
     const match = html.match(/<html\b[^>]*\bstyle="([^"]*)"/i);
-    return match ? match[1] : '';
+    return match?.[1] ?? '';
   })();
 
   it('does not clamp the root <html> width to 360px', () => {
--- a/src/sidepanel.html.test.ts
+++ b/src/sidepanel.html.test.ts
@@ -1,0 +1,35 @@
+/**
+ * Regression guard for issue #336.
+ *
+ * The side-panel document (`public/sidepanel.html`) must tolerate host
+ * viewports narrower than 360px. Brave's side panel renders below ~360px
+ * (its icon rail eats width), so a hard `min-width: 360px` on the root
+ * `<html>` forces the document wider than the viewport and clips the right
+ * edge (balance glyphs, the Faucet button, token fiat values) with no
+ * horizontal scrollbar — the content becomes unreachable.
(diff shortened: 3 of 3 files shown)


0xnullifier (approved):
LGTM

PR #367, opened 2026-07-22
fix: requeue guardian transactions on a still-pending 409 instead of failing (#335)

### LINKED ISSUE #335: Guardian 409 conflict_pending_delta: fixed 12-retry burn (~65 s) then terminal Failed — should back off and requeue
### Problem
When the guardian answers a delta/proposal push with `409 conflict_pending_delta` (another candidate is pending on the account), the wallet retries exactly 12× (~5 s apart, ~65 s total inside guardian-approving) and then marks the transaction **permanently Failed**.

The conflict is transient guardian-side state (see OpenZeppelin/guardian#312): the account is locked for at most grace+retries (~18 min; ~20 s for the diverged case since guardian v0.15.1) and then released. Failing terminally after 65 s turns a temporary lock into a user-visible failed payment.

Evidence from a 300-send stress run (guardian v0.15.1, testnet, 2026-07-14): 56 sends failed with exactly this signature — every one burned exactly 12 in-window 409s (min=median=max=12, i.e. deterministic budget exhaustion, not flakiness); 5 further sends that caught 3–9 conflicts survived and formed the run's p99 latency tail (37–81 s).

### Proposed fix
1. Treat `conflict_pending_delta` as retryable-later: return the tx to Queued with exponential backoff instead of marking Failed.
2. Once OpenZeppelin/guardian#304 lands, key this on the machine-readable `meta.retryable` / `retry_after_secs` instead of matching error strings.
3. Related second gap: when a tx dies locally *after* guardian approval (prover timeout, RPC submit error), the approved candidate stays stranded guardian-side and 409-locks the account for ~18 min. Proactively releasing it needs the abandon endpoint requested in OpenZeppelin/guardian#319.
4. Dependency: OpenZeppelin/guardian#314 (merged 2026-07-14) fixes an ECDSA signer bug that creates exactly these stranded candidates — bump the guardian SDK packages when the release publishes.

### References
OpenZeppelin/guardian#312 · OpenZeppelin/guardian#304 · OpenZeppelin/guardian#313 / OpenZeppelin/guardian#314

WiktorStarczewski:
## Summary

A guardian transaction that receives a still-pending `409 ConflictPendingDelta` after exhausting the conflict-retry budget was being marked terminally `Failed`. Because a pending delta clears on its own once the prior delta canonicalizes, the transaction should have been returned to the queue for a later cycle instead of failing permanently. This change requeues those transactions so they retry automatically.

## Root cause

The guardian holds a single-delta lock. A proposal issued while a prior delta is still canonicalizing returns `409 ConflictPendingDelta`. `withGuardianConflictRetry` retries within a bounded budget, but if the lock does not clear in time the error propagates out of the creator. In `generateTransaction`, any error that was not otherwise handled fell through to `cancelTransaction`, which marks the transaction `Failed`. A still-pending conflict is transient, not terminal, so this produced a spurious permanent failure for a transaction that would have succeeded moments later.

## Fix

In `src/lib/miden/transaction/index.ts`, before the terminal `cancelTransaction` fallback, detect a still-pending guardian conflict via `isGuardianPendingConflict(error)`. When matched:

- reset the status to `Queued` and clear `processingStartedAt` (a bare return would leave it in `GeneratingTransaction`, which `cancelStuckTransactions` would then reap as stalled),
- set the stage back to `creating-proposal`,
- return so the next `generateTransactionsLoop` cycle retries it.

The retry-wrapped creators are side-effect-free, so requeueing is safe. Non-transient 409s (for example a paused account) are not matched by `isGuardianPendingConflict` and continue to fail fast, and `cancelStaleQueuedTransactions` (`MAX_QUEUED_AGE`) remains the terminal cap so a transaction cannot loop forever.

## Testing

- `npx jest src/lib/miden/transaction/transactions.guardian.test.ts --no-cache --runInBand` — 31 passed. Added two cases: a still-pending 409 requeues (status back to `Queued`, `processingStartedAt` cleared, nothing signed or submitted) and a paused-account 409 still fails fast.
- `npx eslint src/lib/miden/transaction/index.ts src/lib/miden/transaction/transactions.guardian.test.ts --max-warnings 0` — clean.

Closes #335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.15.9 (TBD)
+
+### Fixes
+
+* [FIX][all] **Guardian transactions that hit a transient pending-delta conflict now requeue and retry instead of failing.** A guardian `409 conflict_pending_delta` that outlasted the ~60s inline retry budget fell through to a terminal `Failed`, even though the conflict is transient (a single in-flight delta clears on its own). Such transactions are now reset to `Queued` (with `processingStartedAt` cleared) so the processing loop retries them, bounded by the existing 30-minute queued-age cap.
+
 ## 1.15.8 (2026-07-21)
 
 ### Features
--- a/src/lib/miden/back/transaction-processor.ts
+++ b/src/lib/miden/back/transaction-processor.ts
@@ -153,7 +153,11 @@ export async function startTransactionProcessing(): Promise<void> {
     }
 
     let attempts = 0;
-    const maxAttempts = 60; // Max 5 minutes (60 * 5 seconds)
+    // Loop-pass ceiling, not a wall-clock bound: a single pass can now spend up
+    // to the guardian conflict-retry budget (~60s), so 60 passes is NOT "5
+    // minutes". Terminal per-tx caps live elsewhere (MAX_QUEUED_AGE /
+    // MAX_WAIT_BEFORE_CANCEL).
+    const maxAttempts = 60;
 
     while (attempts < maxAttempts) {
       attempts++;
--- a/src/lib/miden/db/types.ts
+++ b/src/lib/miden/db/types.ts
@@ -107,6 +107,15 @@ export interface ITransaction {
    * `status`, and is stale once `status` reaches `Completed`/`Failed`.
    */
   stage?: ITransactionStage;
+  /**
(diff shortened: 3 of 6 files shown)


0xnullifier (approved):
LGTM, A note to be taken is this may not be needed after https://github.com/OpenZeppelin/guardian/issues/337

PR #368, opened 2026-07-22
fix: serialize note import queue read-modify-write to prevent dropped notes (#333)

### LINKED ISSUE #333: Note import queue has unpatched TOCTOU race condition
## Packages versions
miden-wallet: latest main (commit 0294ca2)
@miden-sdk/miden-sdk: v0.15.x

## Bug description
In `src/lib/miden/activity/notes.ts`, there is a Time-Of-Check-Time-Of-Use (TOCTOU) race condition between `queueNoteImport()` (which appends to the import queue) and `importAllNotes()` (which rewrites the queue after processing). The code explicitly acknowledges this race in comments (lines 66-71), but it remains unpatched.

A concurrent `queueNoteImport()` call between the read and write in `importAllNotes()` can cause:
1. A newly queued note to be lost (overwritten by the rewrite)
2. A processed note to be re-added to the queue (wasted retries)

For private notes, which only exist in the queue and are not recoverable from the network, a lost note means **permanent funds loss** after `MAX_IMPORT_ATTEMPTS` (3 attempts) are exhausted.

Potentially related to #250, #260, #56.

## Relevant code
```typescript
// src/lib/miden/activity/notes.ts ~line 66-71
// NOTE: queueNoteImport's read-modify-write is not serialized against this
// rewrite (it runs outside the WASM lock), so a concurrent enqueue landing
// between the read and write below can be lost or a processed note re-added.
// Neither can re-brick the wallet (a re-added note is just retried under the
// cap), but a queue-level lock shared by queueNoteImport and this rewrite
// would close the window entirely.
```

## How can this be reproduced?
See issue body for full reproduction steps including unit test, real-world scenario, and stress test.

## Expected behavior
Queue operations should be serialized. No notes should be lost during concurrent enqueue/import.

## Suggested fix
Add a shared mutex for all queue operations. See issue body for full implementation.

## Relevant log output
```
Dropping queued note after 3 failed import attempts (1234 b64 chars)
```
Note: This log appears after the race has already caused note loss.

WiktorStarczewski:
## Summary

The pending note-import queue (`miden-notes-pending-import`) was read-modified-written from two code paths with no serialization between them. Under concurrency this created a time-of-check-to-time-of-use (TOCTOU) race that could silently drop a queued private note — whose serialized bytes are its only copy — or re-add an already-processed note.

## Root cause

Both `queueNoteImport` and the end-of-pass rewrite in `importAllNotes` perform an independent read-modify-write of the same storage key:

- `queueNoteImport` reads the queue, appends the new note, and writes it back.
- `importAllNotes` rebuilds the queue as `[...retry, ...appendedDuringPass]`, where `appendedDuringPass` is a re-read of the queue sliced past the snapshot it processed.

Nothing serialized these two sequences. If an enqueue landed between `importAllNotes`' rewrite read and its rewrite write, the rewrite wrote a stale array back over the top of the freshly enqueued note, clobbering it. Because a private note's bytes are its only copy, the note was lost entirely; the mirror case re-added a processed note.

## Fix

Introduce a lightweight promise-chain mutex, `withQueueLock`, and route every read-modify-write of the queue key through it — both `queueNoteImport` and the `importAllNotes` rewrite now hold the same lock, so their storage sequences can no longer interleave.

The lock is intentionally scoped to only the storage read/slice/write. It does not cover the WASM `importNoteBytes` work or the post-rewrite `syncState` delay, so an enqueue is never blocked for longer than a single storage round-trip.

## Testing

- Added a regression test that parks `importAllNotes` at its rewrite write, enqueues a fresh private note into the read-write window, then releases the gate and asserts the enqueued note survives the pass.
- `npx jest src/lib/miden/activity/notes.test.ts --no-cache --runInBand` — 10/10 passing.
- `npx eslint src/lib/miden/activity/notes.ts src/lib/miden/activity/notes.test.ts --max-warnings 0` — clean.

Closes #333


0xnullifier (commented):
I don't think this is even needed as there is no longer a way to import notes in the app

WiktorStarczewski:
the manual import-note UI is gone yeah (removed in a5b96be1 + the v0 revamp #248), but that was a separate path that called `importNoteBytes()` directly — never touched this queue.

this code is the dApp `consume` path: a dApp sends a consume with `noteBytes` → `requestConsumeTransaction` queues it (`dapp.ts:1405`) → `importAllNotes` drains it in `generateTransactionsLoop`. still fully wired — the SDK exposes `Transaction.createConsumeTransaction(..., noteBytes)` and there's a test at `dapp.extension.test.ts:974` asserting `queueNoteImport` fires.

so not dead. and for private notes the bytes are the only copy, so the race can actually drop funds → keeping the fix.

PR #369, opened 2026-07-22
fix: recover Dexie and surface stuck-tx heal failures (#254)

### LINKED ISSUE #254: Stuck-tx heal alarm silently swallows errors and never re-opens Dexie — no diagnostic surface
## Background

The SW-side `STUCK_TX_HEAL_ALARM` self-heal mechanism in [`src/lib/miden/back/transaction-processor.ts#L147-L155`](../blob/main/src/lib/miden/back/transaction-processor.ts#L147-L155) catches alarm-handler errors with a bare `console.warn` and continues. That defeats the whole point of a defense-in-depth path: when the heal fails, nothing escalates.

```ts
} else if (alarm.name === STUCK_TX_HEAL_ALARM) {
  cancelStuckTransactions().catch(err =>
    console.warn('[TransactionProcessor] Stuck-tx heal alarm error:', err)
  );
}
```

A pioneer hit this on Chrome MV3 with `DatabaseClosedError: Database has been closed` from Dexie inside `cancelStuckTransactions` → `Repo.transactions.toArray()` ([`activity/transactions.ts:550-562`](../blob/main/src/lib/miden/activity/transactions.ts#L550-L562)). The error fired repeatedly, the heal never recovered, and there was no surface in the UI to notice or react. They only saw it by opening DevTools.

## Problem

1. The heal handler's only failure signal is a `console.warn` — invisible unless DevTools is open.
2. The handler doesn't attempt to re-open / re-acquire the Dexie connection before retrying. If the SW respawn left a stale handle inside the WASM client or any Dexie reference, every subsequent firing of the alarm hits the same error.
3. There is no metric, no Sentry / telemetry hook, no user-visible state — so a wallet stuck in this state looks like normal idle from outside.

## Proposed Solution

Two small, independent changes:

- `cancelStuckTransactions().catch(...)` should at minimum write the error to a persistent surface (sentry, an internal error log table, or the same persisted-warning channel the rest of the wallet uses for SW-side problems) so that "heal silently failing for 4 hours" can be diagnosed without DevTools.
- On `DatabaseClosedError` specifically, the handler should attempt a re-open before giving up. Either by re-importing the `db` module (forcing a fresh `new Dexie('TridentMain')` reference) or by calling whatever the wallet's standard "ensure Dexie is open" helper is.

Any thoughts on this are highly appreciated — happy to send a PR if the direction's right.

## Related (differentiated)

- #250 (sticky in-memory claiming flag) — different layer (Zustand front-store) but same theme: a self-heal mechanism that can't observe its own failure mode.
- #251 (stuck consuming private notes from dApp) — surfaces in a similar way (visible "stuck" state with no recovery), different root cause.

WiktorStarczewski:
## Summary

The stuck-transaction self-heal (`cancelStuckTransactions`) runs both on service-worker startup and on the `miden-tx-stuck-heal` alarm. Both call sites attached a bare `.catch(err => console.warn(...))`. That handler swallowed every failure with no recovery and no field-visible diagnostic, so a wedged heal could recur silently on every alarm tick forever.

The most common wedge is a stale Dexie handle: an MV3 service-worker respawn can leave the `TridentMain` connection closed, so the first read inside `cancelStuckTransactions` rejects with `DatabaseClosedError`. With the old handler, every subsequent alarm tick failed identically and the heal never made progress.

## Root cause

- `console.warn` output is invisible without an attached DevTools session, so a persistently-failing heal was undiagnosable in the field.
- Neither call site attempted to recover the Dexie connection, so a closed handle turned a transient condition into a permanent one — the heal never re-opened the DB.

## Fix

Introduce a single `healStuckTransactions()` wrapper used by both the alarm listener and the startup path:

- On `DatabaseClosedError`, re-open the Dexie connection once (`Repo.db.open()`) and retry the heal. This recovers the common MV3 respawn case without escalation.
- If the retry still fails, or the original error is not a `DatabaseClosedError`, escalate via `reportHealFailure()` instead of swallowing it.
- `reportHealFailure()` logs `console.error` and additionally surfaces a `StuckTxHealFailed` event on the shipped telemetry sink (Segment). The analytics module is imported lazily and guarded exactly like `getBrowser()` — it throws at module load when the Segment write key is absent and pulls in a Node SDK, so it is only ever evaluated inside this service-worker-only error branch, and reporting failures are themselves swallowed so telemetry can never break the heal.

## Testing

- `npx jest src/lib/miden/back/transaction-processor.test.ts --no-cache --runInBand` — 19 passed. New cases cover: re-open + retry succeeds on `DatabaseClosedError` (no escalation), persistent failure after re-open escalates via `trackEvent`, and a non-`DatabaseClosedError` escalates without attempting a re-open.
- `npx eslint src/lib/miden/back/transaction-processor.ts src/lib/miden/back/transaction-processor.test.ts --max-warnings 0` — clean.

Closes #254

### DIFF
--- a/src/lib/miden/back/transaction-processor.test.ts
+++ b/src/lib/miden/back/transaction-processor.test.ts
@@ -16,6 +16,8 @@
 const mockAlarmsCreate = jest.fn();
 const mockAlarmsClear = jest.fn();
 const mockAlarmsOnAlarm = { addListener: jest.fn() };
+const mockStorageGet = jest.fn();
+const mockStorageSet = jest.fn();
 
 // The real webextension-polyfill module shape. Tests override the
 // import behavior in specific cases to force rejection.
@@ -24,6 +26,12 @@ const mockPolyfill = {
     create: (...args: unknown[]) => mockAlarmsCreate(...args),
     clear: (...args: unknown[]) => mockAlarmsClear(...args),
     onAlarm: mockAlarmsOnAlarm
+  },
+  storage: {
+    local: {
+      get: (...args: unknown[]) => mockStorageGet(...args),
+      set: (...args: unknown[]) => mockStorageSet(...args)
+    }
   }
 };
 
@@ -48,11 +56,6 @@ jest.mock('lib/miden/repo', () => ({
   db: { open: (...args: unknown[]) => mockDbOpen(...args) }
 }));
 
-const mockTrackEvent = jest.fn();
-jest.mock('./analytics', () => ({
-  trackEvent: (...args: unknown[]) => mockTrackEvent(...args)
-}));
-
 const mockWithUnlocked = jest.fn();
 jest.mock('./store', () => ({
   withUnlocked: (fn: (ctx: unknown) => unknown) => mockWithUnlocked(fn)
@@ -70,7 +73,8 @@ beforeEach(() => {
   mockSafeGenerateTransactionsLoop.mockResolvedValue({ success: true });
   mockCancelStuckTransactions.mockResolvedValue(undefined);
   mockDbOpen.mockResolvedValue(undefined);
-  mockTrackEvent.mockResolvedValue(undefined);
+  mockStorageGet.mockResolvedValue({});
+  mockStorageSet.mockResolvedValue(undefined);
 });
 
 /** Flush a few microtask / macrotask ticks so in-flight awaits can progress. */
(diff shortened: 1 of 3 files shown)


src/lib/miden/back/transaction-processor.test.ts
```diff
@@ -217,6 +236,149 @@ describe('setupTransactionProcessor', () => {
...
+
+    expect(mockDbOpen).toHaveBeenCalledTimes(1);
+    expect(mockCancelStuckTransactions).toHaveBeenCalledTimes(2);
+    // A recovered heal must not escalate — no diagnostic gets persisted.
+    expect(mockStorageSet).not.toHaveBeenCalled();
+  });
+
+  it('persists a diagnostic to chrome.storage.local when the heal keeps failing after re-open (issue #254)', async () => {
+    // The Segment escalation the PR originally used is dead in the shipped SW
+    // build (`back/analytics.ts` throws without ALEO_WALLET_SEGMENT_WRITE_KEY,
```
0xnullifier, line 264: yeah this reminds we should remove all the ALEO naming related stuff in the codebase right ?

0xnullifier (approved):
LGTM

src/lib/miden/back/transaction-processor.test.ts
WiktorStarczewski: Yeah that would be nice

PR #370, opened 2026-07-22
fix: account-scope extension consumable-notes cache to prevent wrong-account auto-consume (#280)

### LINKED ISSUE #280: [extension/multi-account] Stale consumable notes cache served across account switch — auto-consume fires on wrong account without user action
## Version

- Wallet SDK: `@miden-sdk/miden-sdk 0.15.0-alpha.7` / `@miden-sdk/react 0.15.0-alpha.7`
- Branch/commit: `87e0a7f1`
- Build: extension
- Network: devnet

## Starting state

Multi-account session — two accounts (public + private) in the same wallet profile.

## Steps

1. Account A (private) is the active account — SW syncs and writes Account A's consumable notes to `chrome.storage.local`
2. Switch to Account B (public) before the next sync fires
3. Account B's Explore screen shows Account A's note as claimable
4. Auto-consume fires automatically — no user click needed
5. `TridentMain.transactions` receives a new row with Account B's `accountId` and Account A's `noteId`, status → Failed

## Expected

After switching to Account B, only Account B's own consumable notes should appear. Account A's notes should not be visible or auto-consumed under Account B.

## Observed (facts only)

Both accounts' consumable notes appeared in both account views simultaneously. Account B (public) initiated a consume transaction for a note belonging to Account A (private). The protocol rejected it with "Error: failed to…".

## Does NOT reproduce on

- Build: mobile/desktop (the local path `useLocalClaimableNotes` correctly scopes by account address)
- Single-account session

## Evidence

`TridentMain.transactions` dump, devnet session 2026-06-18:

| accountId | status | initiated | completed | stuckFor | noteId |
|---|---|---|---|---|---|
| PRIVATE `mdev1apkqy0z3...` | Completed | 2:18:0x PM | 2:41:01 PM | 23 min | `0x2e624cf74fe732` |
| PUBLIC `mdev1az9s26jdh...` | Failed | 2:18:5x PM | 2:41:10 PM | 22 min | `0x2e624cf74fe732` |

Same `noteId` in two rows under two different accounts. PUBLIC/Failed completed 9 seconds after PRIVATE/Completed — consistent with the protocol rejecting a consume of an already-consumed note.

`MidenClientDB_mdev.inputNotes`: all 10 entries show `accountId: undefined` — the WASM client stores notes in a shared table without an account column; filtering is handled internally by the Rust layer.

## Post-bug actions taken

Notes eventually consumed successfully after the Failed attempts timed out (~22–30 min, matching `MAX_WAIT_BEFORE_CANCEL`). No manual recovery steps were needed beyond waiting.

## Root cause

Two-layer gap in the extension note-caching path:

**Layer 1 — `sync-manager.ts:244–246`:** The SW writes `miden_cached_consumable_notes` to `chrome.storage.local` as a single wallet-wide key with no account attribution. Account switching does not trigger a re-sync, so the previous account's notes persist in the cache until the next alarm fires.

**Layer 2 — `claimable-notes.ts:175`:** `useExtensionClaimableNotes` accepts a `_publicAddress` parameter (underscore = intentionally unused since `82f1a793` / PR #178) but never reads it. All account views are served the same wallet-wide `extensionClaimableNotes` Zustand slot with no account filter. The `extensionClaimableNotes` slot is also not cleared on account switch — `resetSeenNotes()` (called on switch) only resets toast state.

Note: a correctly-scoped pattern already exists — `miden_sync_data` (the parallel key) embeds `accountPublicKey` and is read with an account-equality check in `useIntercomSync.ts:63` for balance rehydration. The same guard is missing from the notes path.

## Fix direction

- Scope `miden_cached_consumable_notes` by account so each account reads only its own cached notes
- Clear `extensionClaimableNotes` in Zustand immediately on account switch
- Wire up the `_publicAddress` parameter in `useExtensionClaimableNotes` to filter notes before serving them, matching the existing pattern in `useIntercomSync.ts`

## Impact

- **Severity:** Feature degraded — wrong-account Failed consume rows accumulate in `TridentMain.transactions` and compound #250 (sticky claiming flag) and #260 (mutex pressure)
- No funds are lost — the protocol correctly rejects the wrong-account consume

## Notes / hypotheses (speculation)

The stuck duration (22–30 min) also matches the #260 mutex cascade class. The two concurrent wrong-account + right-account consumes for the same note may have added queue pressure that contributed to both notes appearing stuck simultaneously.

## Related

- #250 — Sticky `extensionClaimingNoteIds`: wrong-account Failed consume feeds this directly
- #260 — Recurring stuck Consuming notes: concurrent wrong-account tx adds mutex pressure
- #174 — Wallet adapter account switching: different subsystem, same account-scoping class

WiktorStarczewski:
## Summary

Fixes a cross-account cache leak in the extension's consumable-notes polling. After switching accounts, the previously selected account's cached notes could be served to — and auto-consumed under — the newly selected account.

## Root cause

`useExtensionClaimableNotes` polled `chrome.storage.local` for the wallet-wide `miden_cached_consumable_notes` key. That key is not account-scoped: the service worker overwrites it on every sync cycle regardless of which account the notes belong to. Between an account switch and the next sync tick, the poll would read the outgoing account's notes and push them into the store for the incoming account, which then displayed and auto-consumed them.

The hook also ignored its `publicAddress` argument entirely (`_publicAddress`), so there was no account guard at all.

## Fix

- Read the account-scoped `miden_sync_data` key instead. It carries both the `notes` and the `accountPublicKey` they belong to (written alongside the legacy key in `sync-manager.ts`), and is the same channel `useIntercomSync` already consumes.
- Only push notes into the store when `syncData.accountPublicKey === publicAddress`; otherwise clear to `[]` so a stale set is never shown or auto-consumed.
- When no sync data has been cached yet, leave the store untouched so `isLoading` stays true rather than prematurely flipping to an empty result.
- Add `publicAddress` to the effect's dependency array so the poll re-scopes on account switch.
- Clear `extensionClaimableNotes` in `resetSeenNotes` (invoked on account switch) so the previous account's notes are dropped immediately; the account-scoped poll repopulates them for the new account on its next tick.

## Testing

- `claimable-notes.test.tsx`: added cases proving cached notes are served only to the matching account, and that the poll stays loading (does not overwrite) when no sync data exists yet.
- `store/index.test.ts`: added a case proving `resetSeenNotes` clears stale extension claimable notes.
- Full suites pass: `claimable-notes.test.tsx`, `store/index.test.ts`, `sync-manager.test.ts`. Lint clean on all changed files.

Closes #280


0xnullifier (approved):
LGTM, but I think this was an old issue as we only have single account not but anyways will help in future

PR #372, opened 2026-07-22
fix: validate BIP-39 checksum on seed-phrase restore (#134)

### LINKED ISSUE #134: Wallet Restoration via Seed Phrase Succeeds Using Repeated Single Word Input

WiktorStarczewski:
## Summary

Restoring a wallet from a seed phrase accepted any 12 words that happened to be in the BIP-39 wordlist, even when the phrase failed its BIP-39 checksum (for example a repeated word such as `abandon abandon ... abandon`). Such input is not a valid mnemonic: instead of restoring the user's existing wallet it silently derived an unrelated one. This change enforces the checksum before the restore can be submitted.

## Root cause

`ImportSeedPhraseScreen` computed validity as `seedPhrase.every(word => wordslist.includes(word))` — pure per-word membership. Membership is necessary but not sufficient: a valid BIP-39 mnemonic also encodes a checksum in its final word, so many wordlist-only-valid combinations are not real mnemonics. The submit button was enabled and `onSubmit` fired for these checksum-invalid phrases.

## Fix

- Split the old `isValid` into `allWordsKnown` (per-word wordlist membership) and a stricter `isValid` that additionally requires `validateMnemonic(formatMnemonic(...))` from `bip39` to pass.
- Gate submit (button `disabled` state and the `handleSubmit` guard) on the stricter `isValid`, so a checksum-invalid phrase can never be submitted.
- Surface a distinct error (`justValidPreGeneratedMnemonic`) when every word is known but the checksum fails, so the user is not left with an enabled-looking form or the misleading per-word typo error.

## Testing

- `npx jest src/screens/onboarding/import-wallet-flow/ImportSeedPhrase.test.tsx --no-cache --runInBand` — 12 passed. Added cases cover: a checksum-invalid phrase whose words are all known keeps submit disabled and never emits; the same input surfaces the distinct checksum error rather than the wordlist-typo error; the canonical valid BIP-39 vector (`11x abandon + about`) still enables submit and emits the joined phrase.
- `npx eslint <changed files> --max-warnings 0` — clean.

Closes #134


src/screens/onboarding/import-wallet-flow/ImportSeedPhrase.tsx
```diff
@@ -34,7 +36,20 @@ export const ImportSeedPhraseScreen: React.FC<ImportSeedPhraseScreenProps> = ({
...
+  // from the wordlist (e.g. a repeated word) would pass and silently derive an
+  // unrelated wallet instead of restoring the user's existing one.
+  const isValid = useMemo(
+    () => allWordsKnown && validateMnemonic(formatMnemonic(seedPhrase.join(' '))),
+    [allWordsKnown, seedPhrase]
+  );
+
+  // Distinguish a per-word typo (word not in wordlist) from a checksum mismatch
+  // (every word is known but the phrase is not a valid mnemonic).
+  const isChecksumError = allWordsKnown && !isValid;
```
0xnullifier: LIttle weird to look at no ?
What would look better is
```typescript
const isChecksumValid = useMemo(
    () => validateMnemonic(formatMnemonic(seedPhrase.join(' '))),
    [seedPhrase]
  );

const isValid = isCheckSumValid && allWordsKnown
```

0xnullifier (approved):
LGTM, I know we aren't strict about code quality now with AI but I feel like we should be

PR #374, opened 2026-07-22
fix: stop double-adding chain height to note reclaim height (#308)

### LINKED ISSUE #308: Recall Height is double-added: on-chain reclaim height lands at ~2× chain height, recall fails for days (all notes, public + private)
## Summary

Recall Height is broken for every note that uses it, public and private alike. The Send screen converts the chosen recall date into an absolute block height, but the SDK-interface layer treats that value as a relative offset and adds the current block height a second time before building the transaction. The result is an on-chain P2IDE reclaim height of roughly twice the chain height at send time — with testnet at ~278k blocks, a "30 minutes" recall becomes recallable ~9.6 days later. Every reclaim attempt before then fails with the kernel's (correct) refusal. The chain, node, sync, and note transport all behave correctly; the wallet hands them the wrong number.

Versions: wallet v1.15.3 (extension build, commit tagged v1.15.3), @miden-sdk/miden-sdk 0.15.2, @miden-sdk/react 0.15.2 (verified from package.json). The defect is present in every release from v1.14.3 through v1.15.3 and is still on current main ([src/lib/miden/sdk/miden-client-interface.ts#L478-L481](../blob/main/src/lib/miden/sdk/miden-client-interface.ts#L478-L481)).

## Symptoms

Every sender-side recall attempt on a Recall-Height note fails. Persisted transaction error (identical on all 126 failed attempts observed in one affected wallet):

```
Error: failed to execute transaction: transaction execution failed: failed to execute transaction kernel program:
  x assertion failed with error message: failed to reclaim P2IDE note because the reclaim block height is not reached yet
```

User-visible: the recall date passes, the recall/claim-back action keeps failing with the error above, funds appear stuck. Persists across retries, popup re-opens, and browser restarts (the wrong height is baked into the note on-chain, so nothing client-side can fix an already-sent note).

## Reproduction steps

1. On wallet v1.15.3 (testnet), open Send, choose any token, amount, and recipient.
2. Enable Recall Height and pick any date/time or preset (e.g. "30 mins").
3. Confirm the send (public or private — both are affected).
4. Decode the created note's inputs (third input value is the reclaim block height), or wait past the chosen recall time and attempt to recall the note as the sender.

Expected: the note's reclaim height equals the height the UI computed for the chosen date (current height + ~600 blocks for a 30-min preset), and the sender can reclaim after that time.

Actual: the note's reclaim height equals current-height-at-send PLUS the UI's already-absolute value (~2× chain height, ~9.6 days out at current testnet height), and every earlier reclaim attempt fails with the kernel assertion above.

Reproduction rate: 100% — 15/15 notes (10 private, 5 public) sent from one wallet on 2026-07-03 all carry the doubled height, with the exact identity on-chain-reclaim = height-at-send + UI-value holding on every note.

Live-reproduced on Windows 11 / Chrome 149 / wallet v1.15.3 (testnet) on 2026-07-03, via read-only inspection of the affected wallet's IndexedDB plus live node queries. A fresh instrumented send was not initiated.

## Root cause (2 locations, unit-contract mismatch)

**[src/screens/send-flow/SendDetails.tsx#L25-L29](../blob/main/src/screens/send-flow/SendDetails.tsx#L25-L29)** — dateTimeToRecallBlocks returns an ABSOLUTE height: current block number plus seconds-until-target divided by three. This absolute value is stored on the transaction record (extraInputs.recallBlocks) and passed through unchanged.

**[src/lib/miden/sdk/miden-client-interface.ts#L478-L481](../blob/main/src/lib/miden/sdk/miden-client-interface.ts#L478-L481)** — sendTransaction treats that field as a RELATIVE offset and computes reclaimAfter as current sync height plus extraInputs.recallBlocks — adding the chain height a second time. The SDK's reclaimAfter is documented as an absolute block number and bakes the doubled value verbatim into the P2IDE note inputs.

Provenance: the original send UI (RecallBlocksModal, "Blocks until recall", Nov 2024) collected a relative offset, which matched the interface's addition. The send-flow redesign (first shipped in v1.14.3) switched the UI to the absolute date-picker computation without removing the addition in the interface layer. The dApp send path ([src/lib/miden/back/dapp.ts#L1151-L1160](../blob/main/src/lib/miden/back/dapp.ts#L1151-L1160)) feeds the same field and is equally affected. RecallBlocksModal is now dead code (no imports).

## Receipts (evidence chain)

| Signal | Source | What it proves |
|---|---|---|
| 15 sends with extraInputs.recallBlocks set (10 private, 5 public), values 277,534–277,905 ≈ chain height at send + ~580 blocks | Affected wallet's TridentMain.transactions (read-only dump, 2026-07-03) | UI stored an absolute height; user intent (~30-min preset) captured correctly at UI stage |
| Decoded P2IDE note inputs: reclaim height = height-at-send + recallBlocks EXACTLY on all 13 decodable notes (e.g. 276,967 + 277,534 = 554,501); timelock = 0 | MidenClientDB_mtst.inputNotes inputs field, cross-joined with the send records | The double-add materialized in the actual notes |
| Node-side note fetch for public notes returns the same doubled heights (555,153 / 555,229) in the note inputs | Live rpc.testnet.miden.io GetNotesById, 2026-07-03 | The wrong height is on-chain, from an independent witness — not just wallet-local records |
| 126 consume attempts on the 15 notes, all Failed with "failed to reclaim P2IDE note because the reclaim block height is not reached yet"; attempts continued ~30+ min past every intended recall time | TridentMain.transactions error fields + timestamps | Failure is the on-chain script correctly rejecting early reclaim; not sync, prover, transport, or user impatience |
| Error string and input layout (target, reclaim height, timelock) match the P2IDE note script constants and storage doc | miden-base p2ide.masm (ERR_P2IDE_RECLAIM_HEIGHT_NOT_REACHED) | Confirms decode layout and which assertion fires |
| Live nullifier check: 13/15 notes unspent at chain tip 278,518; 2 public notes consumed by the recipient (blocks 277,289 and 278,010) | Live rpc.testnet.miden.io SyncNullifiers | Current status of every note; recipient consumption unaffected by the bug |
| All 10 private note payloads present on the note transport under the recipient's tag | Live transport.miden.io FetchNotes (matched by detailsCommitment + serialNumber) | Note transport is not the failure; delivery worked |
| Relative "Blocks until recall" UI replaced by absolute date-picker computation while the interface addition remained; present in all tags v1.14.3 → v1.15.3 and main | git history (pickaxe on the recall computation and the interface line) | Regression provenance and affected-version range |

## Suggested direction

Single-source the unit and make it impossible to drift again: either pass the UI's absolute height through to the SDK unchanged (dropping the re-addition in the interface layer), or carry a relative offset end-to-end and convert to absolute in exactly one place immediately before the SDK call — and rename the field so its unit is explicit in the type, since this regression was born from a UI rewrite changing semantics while the interface kept the old contract. Audit the dApp send path in the same pass, as it feeds the same field with unspecified semantics. Complementary hardening in the same area: gate the sender-side reclaim action on the note's reclaim height versus current height (show "recallable at ~date" instead of letting the kernel assert), reconcile the "funds will be returned to you" copy with the fact that reclaim is a manual action today, and add an end-to-end test that decodes a freshly built note's inputs and asserts the baked reclaim height equals the UI-intended height for both public and private notes — that single assertion catches this whole bug class. Existing mis-labeled notes cannot be rewritten (heights are immutable in the notes); they remain claimable by recipients anytime and recallable by senders once the erroneous height is reached.

WiktorStarczewski:
## Summary

Recall (P2IDE reclaim) height was being stored on-chain at roughly twice the current chain height, so every sender-side reclaim attempt before that far-future height failed the kernel's `ERR_P2IDE_RECLAIM_HEIGHT_NOT_REACHED` assertion. This makes the "expiration date" the user picks in the send flow unusable — the note only becomes reclaimable at ~2x chain height instead of the chosen offset.

## Root cause

The current chain height was added twice:

- The send UI's `dateTimeToRecallBlocks` returned an **absolute** block height (`currentBlockNum + secondsUntilTarget / SECONDS_PER_BLOCK`), fetching the current height itself.
- `MidenClientInterface.sendTransaction` treats `extraInputs.recallBlocks` as a **relative** offset and adds the current sync height again (`reclaimAfter = sync.blockNum() + recallBlocks`).

The SDK bakes `reclaimAfter` verbatim as the note's absolute P2IDE reclaim height, so the on-chain value landed at `height + (height + offset)` ≈ 2x the chain height. The dApp path already passed `recallBlocks` as a relative offset, so it was unaffected.

## Fix

Single-source the unit so the current height is added in exactly one place:

- `dateTimeToRecallBlocks` now returns a **relative** block offset (0 for a past/now target) and no longer takes/needs a `currentBlockNum`.
- `MidenClientInterface.sendTransaction` remains the one place that adds the current sync height. `SendTransaction.extraInputs.recallBlocks` is documented as a relative offset.
- The now-unnecessary on-mount block-height fetches (RpcClient / `ensureSdkWasmReady`) are removed from `RecallCalendarDrawer` and `ReviewTransaction`, along with their associated effect-lifecycle plumbing.

Net result: the on-chain reclaim height is `syncHeight + offset` instead of `~2 * syncHeight`.

## Testing

- Added a regression guard in `miden-client-interface.test.ts` asserting the interface adds the current chain height exactly once (`reclaimAfter === 1600` for `H = 1000`, `offset = 600`, not `2600`) and syncs once.
- Updated `RecallCalendarDrawer` and `ReviewTransaction` tests to reflect the relative-offset contract and the removal of the block-height fetch.
- All three affected suites pass (96 tests); `eslint --max-warnings 0` is clean on the changed files.

Closes #308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.15.9 (TBD)
+
+### Fixes
+
+* [FIX][all] **Scheduled recall/reclaim height is no longer double-counted.** The send UI produced an absolute reclaim block height while the SDK interface layer added the current chain height again, so a note's on-chain reclaim height landed at roughly twice the chain height and the recall stayed un-executable for days. The UI now emits a relative offset and the interface layer resolves it against the current height exactly once.
+
 ## 1.15.8 (2026-07-21)
 
 ### Features
--- a/src/lib/miden/back/dapp.branches.test.ts
+++ b/src/lib/miden/back/dapp.branches.test.ts
@@ -383,6 +383,25 @@ describe('requestSendTransaction — mobile error branches', () => {
       transaction: { ...validTx, recallBlocks: 100 }
     } as never);
     expect(res.type).toBe(MidenDAppMessageType.SendTransactionResponse);
+    expect(mockRequestConfirmation).toHaveBeenCalledWith(
+      expect.objectContaining({ transactionMessages: expect.arrayContaining(['Recall Blocks, 100']) })
+    );
+  });
+
+  // Regression guard for #308: a 0 recall offset means "reclaimable at the
+  // current height" and IS honored on execution, so the preview must surface
+  // the row (value 0) rather than hide it behind a truthiness check.
+  it('surfaces the recallBlocks row in preview when the offset is 0', async () => {
+    mockInitiateSendTransaction.mockResolvedValue('tx-recall-zero');
+    const res = await dapp.requestSendTransaction('https://miden.xyz', {
+      type: MidenDAppMessageType.SendTransactionRequest,
+      sourcePublicKey: 'miden-account-1',
+      transaction: { ...validTx, recallBlocks: 0 }
+    } as never);
+    expect(res.type).toBe(MidenDAppMessageType.SendTransactionResponse);
+    expect(mockRequestConfirmation).toHaveBeenCalledWith(
+      expect.objectContaining({ transactionMessages: expect.arrayContaining(['Recall Blocks, 0']) })
+    );
   });
 });
 
--- a/src/lib/miden/back/dapp.ts
+++ b/src/lib/miden/back/dapp.ts
@@ -1619,7 +1619,7 @@ function formatSendTransactionPreview(transaction: SendTransaction): string[] {
     `Note Type, ${capitalizeFirstLetter(transaction.noteType)}`
   ];
 
-  if (transaction.recallBlocks) {
+  if (transaction.recallBlocks != null) {
     tsTexts.push(`Recall Blocks, ${transaction.recallBlocks}`);
   }
 
--- a/src/lib/miden/db/types.ts
+++ b/src/lib/miden/db/types.ts
@@ -175,6 +175,10 @@ export class SendTransaction implements ITransaction {
   displayMessage?: string;
   displayIcon: ITransactionIcon;
   delegateTransaction?: boolean;
+  // `recallBlocks` is a RELATIVE block offset (blocks from send time until the
+  // note becomes reclaimable). The absolute P2IDE reclaim height is derived once,
+  // in MidenClientInterface.sendTransaction, as `sync.blockNum() + recallBlocks`.
+  // Do NOT pre-add the current height here or it gets counted twice (see #308).
   extraInputs: { recallBlocks?: number } = {
     recallBlocks: undefined
   };
--- a/src/lib/miden/sdk/miden-client-interface.test.ts
+++ b/src/lib/miden/sdk/miden-client-interface.test.ts
@@ -395,6 +395,51 @@ describe('MidenClientInterface', () => {
 
     expect(result).toBe(fakeTransactionResult);
     expect(fakeMidenClient.transactions.send).toHaveBeenCalled();
+    // No recall requested → the note must be broadcast without a reclaim height.
+    expect(fakeMidenClient.transactions.send).toHaveBeenCalledWith(
+      expect.objectContaining({ reclaimAfter: undefined })
+    );
+    expect(fakeMidenClient.sync).not.toHaveBeenCalled();
+  });
+
+  // Regression guard for #308 (edge case): a RELATIVE offset of 0 means "reclaimable
+  // right now" (the user kept "today" and picked a time that is now/past, so
+  // `dateTimeToRecallBlocks` returned 0). It must resolve to `reclaimAfter = current
+  // height + 0`, NOT be dropped — otherwise a NON-reclaimable note is broadcast and
+  // the sender can never reclaim it.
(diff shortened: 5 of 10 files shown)


WiktorStarczewski:
Superseded by #423 (merged), which fixes the same #308 double-add end-to-end: the send UI's `dateTimeToRecallBlocks` now returns a **relative** blocks-until-recall offset, and `MidenClientInterface.sendTransaction` is the single place that converts it to an absolute height (`reclaimAfter = sync.blockNum() + recallBlocks`). On `main`, `dateTimeToRecallBlocks(targetDate)` is already the relative single-arg version.

One intentional difference: #423 maps a past/now target to `1` rather than `0` — deliberately avoiding `0`, which is falsy in the interface's `if (extraInputs?.recallBlocks)` gate and would read as "no recall." That approach was chosen, so this PR is redundant. Closing.

PR #377, opened 2026-07-22
feat(confirm): human-readable custom-transaction & signature confirmation

WiktorStarczewski:
## Why

Feedback on the wallet: the dApp **custom-transaction** confirmation screen showed only opaque boilerplate — *"This dApp is requesting a custom transaction, please ensure you know the details before proceeding."* plus a single dApp-supplied recipient. As the reviewer put it: *"Is this an arbitrary signature request or a TxSummary? Right now this is extremely sketchy if this were with real assets."*

Every other transaction type already renders friendly details; only **custom** fell through to boilerplate, and the one detail it showed (`recipient`) came from the dApp, not the request bytes.

## What

The wallet now **decodes the request itself** and shows what the transaction actually does — never trusting a dApp-supplied description.

- **Custom transaction → asset summary.** An instant client-side *declared* preview (decoded from the request bytes), then a background **dry-run** (`executeForSummary`, no prove/submit) that replaces it with the **ground-truth vault delta**: what you send / receive, notes consumed/created, storage impact. State machine: `declared → verifying → verified | undecodable`, with an honest *"could not verify"* caveat on any simulation failure.
- **Untrusted data demoted.** The dApp's `recipientAddress` and the raw request move into a collapsible **Advanced** (JSON) disclosure — shown, but never presented as verified truth.
- **Opaque signatures made distinct.** Genuinely-opaque requests (blind `word`, `Arbitrary`, `Blind` signing-inputs) now carry an explicit **⚠ blind-sign warning** + raw value under Advanced, so they can't be mistaken for a decoded transaction.
- **Unverified vs verified amounts are visually distinct** (anti-spoofing): declared figures render muted, verified figures bold red/green.

The existing Sign→`TransactionSummary` asset-diff renderer was extracted into a shared `TransactionAssetView`, so both flows share one component. **The `wallet-adapter` package is untouched** — it already forwards everything the extension needs.

## Key changes

- `src/lib/miden/back/simulate-custom-tx.ts` — new: dry-run (`import notes → syncState → executeForSummary`) under `withWasmClientLock`, time-boxed, never throws (returns `{error}`).
- `src/lib/miden/back/dapp.ts` — custom-tx confirm payload carries the request bytes; `requestConfirm` gains a non-closing `DAppSimulateTransactionRequest` handler.
- `src/lib/miden/types.ts`, `src/lib/store/*`, `src/lib/miden/front/client.ts` — the `simulateCustomTransaction(id)` intercom transport.
- `src/app/confirm/{decode,TransactionAssetView,AdvancedDetails}.{ts,tsx}` — new decode helpers + shared render + collapsible.
- `src/app/ConfirmPage.tsx` — custom-tx state machine + opaque-sign hardening.
- `public/_locales/en/messages.json` — new English copy (other locales auto-generated by CI).

## Testing

- **6323 unit tests pass.** New modules at **100%** coverage; global coverage **97.0% lines / 97.0% statements / 96.3% functions / 97.0% branches** (≥95% gate).
- Built via test-driven, subagent-per-task development with a spec-compliance + quality review after each task and a whole-branch senior review at the end; review findings (dry-run time-box, unverified-amount distinction, hex-address handling, metadata-fetch guard) were applied.

## Notes for reviewers

- **Trust model:** the authoritative asset view is the wallet-executed `executeForSummary` vault delta (extension-internal, un-forgeable). The declared view is always labelled unverified.
- **Pre-confirm dry-run** imports the payload's notes + syncs before the prompt (needed to simulate). This is a local, non-submitting, reversible operation; no auto-consume acts on imported-but-unconsumed notes before confirm.
- **Known follow-up:** a slow `syncState` still holds the single-threaded WASM lock for its duration (the time-box bounds the UI spinner, not the lock). Decoupling sync from the lock is a follow-up.
- **Out of scope:** the mobile/desktop `dappConfirmationStore` confirm path (this PR wires the decoded flow through the extension `requestConfirm` popup), fiat pricing, and orphan-note cleanup after cancel.


src/lib/miden/back/simulate-custom-tx.ts
```diff
@@ -0,0 +1,83 @@
...
+export async function simulateCustomTransaction(input: SimulateCustomTxInput): Promise<SimulateCustomTxResult> {
+  const work: Promise<SimulateCustomTxResult> = (async () => {
+    try {
+      return await withWasmClientLock(async () => {
+        const client = await getMidenClient();
+
+        for (const noteB64 of input.importNotes ?? []) {
+          await client.importNoteBytes(b64ToU8(noteB64));
+        }
+        await client.syncState();
```
0xnullifier, line 68: This runs before the user has approved anything: opening the confirm popup is enough to write dApp-supplied notes into the real client DB, and cancelling doesn't remove them (orphan cleanup is listed as out of scope).

The problem is these notes don't stay invisible. The claimable-notes pipeline — `client.getConsumableNotes(accountPubKey)` in `src/lib/miden/back/sync-manager.ts` (~L156) and `useClaimableNotes` in `src/lib/miden/front/claimable-notes.ts` — has no provenance filter: it lists every tracked input note the account can consume. Notes carried in `importNotes` are by construction consumable by the user's account, so the moment this import + `syncState` completes they:

- appear as ordinary incoming notes in `/pending-notes` (sender + amount rendered like a real payment), and light up the unclaimed-notes badge (`useHasUnclaimedNotes`);
- persist there forever after the user **cancels** the request;
- get consumed — note script and all — if the user later taps **Claim All**, which is exactly the action they declined in the popup. The PR's "no auto-consume touches imported-but-unconsumed notes" claim holds, but the UI actively invites the user to do the consuming manually.

Suggested resolution:
As there is no way to delete notes (I think we need this feature in the web-sdk), let's find a way to mark this notes so that they don't show up in the UI i.e filter the notes in `claimableNotes.ts`

0xnullifier (changes requested):
Requesting changes for one issue:  the pre-confirm dry run introduces notes to the db even if the user cancels the note will be seen in /pending-notes

src/lib/miden/back/simulate-custom-tx.ts
WiktorStarczewski: Good catch, thanks — you're right that the pre-confirm import made those notes claimable, persisted them after cancel, and left them consumable via Claim All.

Fixed with the marking approach you suggested (`0bd423f4`, hardened in `2b8fd3e`):

- A small **quarantine set** of note ids. The dry-run marks the notes it imports, and both claimable read paths filter them out — `sync-manager.ts` (which writes the `miden_cached_consumable_notes` cache the `/pending-notes` list + unclaimed badge read) and `claimable-notes.ts`'s local-client path — so they no longer appear anywhere claimable.
- **Lifecycle:** *confirm* releases the ids (the tx will consume them; if submit fails they reappear, which is correct); *cancel/decline* keeps them hidden.
- The id derivation **mirrors `deserializeNoteFileOrNote`** — it handles both a serialized `NoteFile` (`noteId()`) and a bare `Note` (`.id()`), so a dApp can't dodge the filter by choosing the other serialization format. (Same dual-parse applied to the declared-preview decode so a `NoteFile` swap doesn't blank out.)

Known tradeoffs, both intentional: the set is capped (≥500 pending declines evicts the oldest — not a realistic count for one session), and a *confirmed-but-later-failed* tx keeps its note hidden (the user approved consuming it). Fully agree the real fix is note deletion in the web-sdk once it lands.

Re-requesting review — thanks again for the thorough catch.

PR #382, opened 2026-07-23
fix(mobile): smooth home carousel slide on tab/action tap

WiktorStarczewski:
## What

The programmatic slide between home pages — tapping **Send / Receive / Overview** — felt choppier than a finger swipe between the same pages.

## Why

On the first frame of a *programmatic* slide, WebKit has to create the compositor layer for the `width:500%` carousel track (a full repaint), because the track isn't pre-promoted. A finger drag never pays this: `drag="x"` keeps the track on a live layer the whole time, so it stays smooth.

## Fix

Pre-promote the track to its own compositor layer with `will-change: transform` (`HomeSwipeContainer`). The layer already exists before the slide starts, so a programmatic slide begins on a composited layer instead of triggering layer creation + repaint on frame 1.

### On the reviewed-and-dropped `startTransition` approach

An earlier revision of this PR also wrapped the tab/action `navigate()` in `startTransition` to keep the route-commit re-render from starving the spring's opening frames. Code review found that introduces real regressions, so it was removed:

- **Android hardware-back could minimize the app** — deferring the route re-render leaves `MobileBackBridge` reading a stale pathname/position during the pending window, so a back-press right after a tap falls through to "exit app."
- **Rapid taps to a different action were silently dropped** (the re-tap guards read the stale pathname).
- **The slide could start *later* under sync load** — transitions have no deadline — inverting the goal on exactly the busy devices it targets.

The durable fix for that class — making `useLocation` a `useSyncExternalStore` so navigation state can't lag the URL, after which deferral would be safe — is a separate, larger change, tracked as follow-up rather than blocking this.

## Test plan

- `tsc --noEmit` and `eslint` clean; commit signed.
- **Needs an on-device feel check** (mobile animation): confirm the Overview→Send slide is smoother. If pre-promotion alone isn't enough, the follow-up is the concurrent-safe router migration above — not re-adding `startTransition` as-is.

No web-sdk dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.15.11 (TBD)
+
+### Fixes
+
+* [FIX][mobile] **Tapping a home action (Send/Receive/Overview) now slides between the carousel pages as smoothly as a finger swipe.** The programmatic slide animated the full page width via a main-thread spring whose opening frames were starved by the route-commit re-render, and the track wasn't compositor-promoted so WebKit repainted it as the slide began — while a drag stays on a live layer and only springs a tiny residual. The track is now pre-promoted (`will-change: transform`) and the tab/action navigations defer their route re-render via `startTransition`, so the slide keeps its frames.
+
 ## 1.15.10 (2026-07-23)
 
 ### Changes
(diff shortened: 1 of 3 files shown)


0xnullifier (approved):
LGTM

PR #390, opened 2026-07-27
feat(swap): auto-settle swap order notes

0xnullifier:
Extracts the swap auto-consume / settlement work from `utk-bridge-integration` as a standalone PR off `main`.

## What's included

- **Settlement engine** (`lib/miden/swap/settlement.ts` + `classification.ts`): classifies consumable notes against locally-created PSWAP orders via `pswap.lineage`, queues one batched settlement consume per terminal order (payback claims on fill; expiry reclaim only for orders carrying an explicit `expiresAt` stamp), tags the consume with `swapOrderTxId`/`swapSettleKind`, and self-heals lost settlement stamps.
- **Drivers**: `SwapSettlementManager` (3s tick on mobile/desktop, mounted in `MidenProvider`) and a reconcile pass inside the extension SW's `runSync` (reusing the notes already classified under the WASM lock).
- **Batch consume**: `initiateConsumeNotesTransaction` (one proof/submit for N notes) with per-note dedup over a new multi-entry `noteIds` Dexie index (v1.3 migration + backfill), `manualRetry` bypass of the auto-consume backoff, multi-note support in `MidenClientInterface.consumeNoteId` and the Guardian consume proposal, and summed batch amounts in `completeConsumeTransaction`.
- **History integrity**: settlement consumes are suppressed while their swap row exists (`suppressLinkedConsumes`); token-scoped history matches a swap row on its requested faucet too (`matchesTokenId`); `getSwapSettlementNotes` exposes the settled/reclaimed note ids for detail views.
- **Claim list**: swap-managed auto-consume notes are filtered out of the manual claim UI (local + extension paths) and excluded from new-note notifications, while still entering the seen-note dedup set.

## Also included (swap-flow UI)

- Expiry input and auto-consume toggle on the swap review step, wired through `initiateSwapTransaction`; live quote via the `/v1/swap-eta` endpoint (`useSwapEta`, debounced + supersession-safe) replacing the per-token price fetches, with a "Usually fills in" ETA row and a receive-field loading skeleton; the swap token registry now appends the network's discovered native asset.

## Deliberately not included

- The redesigned history/details UI from the bridge branch (settlement status chip, settlement-note display) — that lands with the history-redesign work on PR #392.

Expiry-reclaim is armed: `completeSwapTransaction` stamps the explicit `extraInputs.expiresAt` (completedAt + expirySeconds) the settlement engine requires — a fix beyond the source branch, where the stamp was missing and the expiry path was dormant. Pre-existing orders without the stamp are still deliberately never treated as expired. The swap-eta host also moved to per-network config (`MIDEN_SWAP_ETA_ENDPOINTS`, devnet fallback).

## Verification

- `tsc --noEmit` clean
- `yarn test:coverage`: 466 suites / 6394 tests pass, coverage 96.9% (gate 95%)


0xnullifier:

PR #391, opened 2026-07-27
feat(guardian): auto-rotate hot key behind blocking overlay + post-rotation 401 fix

0xnullifier:
Extracts the guardian hot-key rotation work from `utk-bridge-integration` as a standalone PR off `main`.

## What's included

- **`HotKeyRotationGate`** (new, mounted in `App.tsx`): full-screen blocking overlay for the current account while it carries `requiresHotKeyRotation` (seed-phrase-recovered / legacy-migrated Guardian accounts with no usable local hot key). Auto-initiates (or adopts) the `replace-hot-key` transaction under a per-account Web Lock so concurrently-open surfaces can't double-rotate, requeues orphaned `GeneratingTransaction` rows only while the processing loop's lock is provably free, drives the FIFO loop on mobile/desktop (SW owns it on extension), swallows hardware back, and offers Retry on failure. Dismissed authoritatively by the flag clearing (`Vault.swapHotKey` → accountsUpdated).
- **Permanent-401 fix** in `completeReplaceHotKeyTransaction`: the guardian's request-auth cosigner allowlist is written only by `/configure` (`registerOnGuardian`), which derives it from the service's in-memory signer set — reusing the pre-rotation service pushed the OLD hot key, so every request signed by the new key 401'd forever. Completion now builds a fresh cold `MultisigService` from the freshly-synced on-chain account and re-registers the POST-rotation signer set (best-effort, before `swapHotKey` arms the hot-sync). The `service` parameter is gone; the call site in the guardian pipeline updated.
- 5 new i18n keys, ported tests (8 gate tests + 2 rebuilt completion tests).

## Deliberately not included

- The wallet-prompts system from the source branch. (`ActivateHotKeyBanner` IS removed here — it is fully superseded by the blocking overlay.) 
- The source branch's guardian pipeline rework (`executeRequest` path) — main keeps its `_withInnerWebClient` path; only the completion call-site signature changed.

## Verification

- `tsc --noEmit` clean; eslint clean
- `yarn test:coverage`: 466 suites / 6387 tests pass, coverage 97.2% statements (gate 95%)


0xnullifier:

PR #393, opened 2026-07-27
feat(prompts): faucet and pending-notes home prompts

0xnullifier:
Extracts the Faucet and Pending-Notes home prompts from `utk-bridge-integration` as a standalone PR off `main` (the Bridge prompt from that branch is excluded — it ships with the bridge).

## What's included

- **Faucet prompt**: shown for a loaded, zero-balance wallet; its inline CTA mints from both the Forkchoice faucet and the official Miden faucet REST API (new `lib/miden-chain/faucet-api.ts`: PoW challenge → solve → `get_tokens`), with loading/success/failure indicators on the card. Auto-seeds when balances settle at zero and auto-completes once a balance appears.
- **Pending-notes prompt**: derived live from claimable notes, showing their USD total (`getPendingNotesUsdTotal`, BigNumber-precise); dismissal is batch-scoped via `pendingNotesDismissedIds` — the prompt resurfaces only when an entirely new batch of notes arrives. CTA navigates to `/pending` (main's current pending surface; the source branch's redesigned PendingNotes page is a separate extraction).
- **PromptCard/PromptCarousel** upgrades from the branch (verbatim): status indicators + inline CTA on the card; callback-ref track, pointer-claim vs the home swipe container, post-drag click suppression, gap-aware paging.
- `formatUsd` i18n helper, `MIDEN_FAUCET_API_ENDPOINTS`, 6 locale keys, Explore→HomeOverview→HomePrompts data plumbing, tests throughout (main's existing PromptCard/Carousel suites kept and extended, not replaced).

## Notes

- **`HotKeyHardwareUnavailable` prompt now included**: mobile-only — when a native hot-key op rejects (device TEE/Secure Enclave unusable), the raw error is recorded and a critical home prompt offers a copy-error action for reporting. Producer call-sites wrap `generateHotKey` (only after the JS fallback also fails), `signHotDigest`, and `revealHotKey` via a lazy import so the secure-hot-key facade stays React-free.
- `ActivateHotKeyBanner` is kept here; PR #391 removes it — whichever merges second resolves a one-line conflict in `HomePrompts`.

## Verification

- `tsc --noEmit` clean; eslint + `lint:i18n` clean
- `yarn test:coverage`: 466 suites / 6421 tests pass, 97.22% statements (gate 95%)


0xnullifier:

PR #412, opened 2026-07-29
Route guardian local proving to the native prover on iOS

WiktorStarczewski:
## Problem

On iOS, switching a **guardian (multisig) account** to **local (non-delegated) proving** froze the UI for ~13s during the prove.

## Root cause

The guardian pipeline (`generateGuardianTransaction`) drives the raw client directly. Its delegated branch was already hardened to use the native prover on mobile, but its **local-proving branch hardcoded the single-threaded WASM prover** with no `isMobile()` check:

```ts
if (!transaction.delegateTransaction) {
  provenTx = await executedTx.prove({ prover: TransactionProver.newLocalProver() }); // WASM, main thread
}
```

On iOS WKWebView, `newLocalProver()` runs WASM single-threaded **on the JS main thread**, blocking the UI for the whole multi-second prove. (Non-guardian paths already route through `proveWithFallback`, which picks the native prover on mobile — so this was guardian-specific.)

## Fix

Route the guardian local branch to the native Rust prover on mobile, mirroring `proveWithFallback`'s `localProverFactory` and the delegated fallback three lines below:

```ts
const localProver = isMobile()
  ? TransactionProver.newCallbackProver(buildNativeProverCallback())  // native, off-thread
  : TransactionProver.newLocalProver();
provenTx = await executedTx.prove({ prover: localProver });
```

## Verification

- **On-device (iPhone 17 Pro Max, iOS 26, testnet):** guardian local prove now goes through `[MidenNativeProver] prove()` and returns in **~795ms off the main thread** (`prove() success, output=91276 bytes`), UI responsive — vs the ~13s main-thread WASM freeze before.
- New regression test pins the mobile branch (`isMobile=true` ⇒ `newCallbackProver` used, `newLocalProver` not); all 40 guardian tests pass.
- Full coverage suite clears the 95% gate (95.37% stmts / 95.38% branches / 95.46% funcs / 95.37% lines).
- Lint clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.15.14 (TBD)
+
+### Fixes
+
+- [FIX][mobile] **Guardian transactions with local (non-delegated) proving no longer freeze the UI on iOS.** The guardian pipeline drives the raw client directly, and its local-proving branch always used the single-threaded WASM prover — which on iOS WKWebView runs on the main thread and locks the UI for the whole multi-second prove (~13s). It now routes to the native Rust prover on mobile (off the main thread), matching the non-guardian path and the guardian delegated-proving fallback.
+
 ## 1.15.13 (TBD)
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


0xnullifier (commented):
LGTM

PR #438, opened 2026-08-02
feat: enable Earn deposits on Guardian accounts

WiktorStarczewski:
## What

Enables Earn (Epoch lending) deposits on **Guardian** (multisig) accounts. Previously the Deposit CTA was disabled with _"Earn deposits aren't available on Guardian accounts yet."_

## Why it was blocked

An Earn collateral note must be a recallable **P2IDE** — it carries a reclaim height that the Epoch allocator validates and that is the user's only escape hatch if the lending leg never settles. The multisig client's send proposal is **P2ID-only**, so the collateral couldn't be expressed on a Guardian account. This is the same limitation that was solved for Guardian *sends* in v1.15.17.

## How

Reuses the v1.15.17 mechanism instead of leaving the feature blocked:

- The Guardian `earn-deposit` branch of `generateTransaction` builds a P2IDE `newSendTransactionRequest` to the Epoch allocator (relative `recallBlocks` → absolute `syncHeight + recallBlocks`, the same relative→absolute conversion the non-Guardian path uses) and proposes it via `createCustomProposal(bytes, 'earn_deposit')`. The request bytes are persisted on the row so a retry after a process restart reuses identical bytes (same rule as send/swap).
- **Completion fix:** the Guardian completion switch now routes `earn-deposit` to `completeEarnDepositTransaction` (extracts the committed collateral note id that `createEarnP2IDNote` hands back to the Epoch SDK). Without this it would have fallen through to the generic custom-tx completion and stranded the deposit.
- Removed the old guard (`GUARDIAN_EARN_DEPOSIT_UNSUPPORTED`), both UI gates (`EarnVaultDetail` / `EarnDepositReview`), the now-dead i18n key, and unused imports.

## Testing

- `yarn test` — full suite (6754 tests) green. Coverage gate clears all four metrics (branches 95.13%, functions 95.09%, lines/statements 95.42%). New tests in `transactions.guardian.test.ts` cover the build-to-allocator, retry-reuse, and missing-`recallBlocks` fail-fast paths; `earn.test.ts` and both earn screen tests updated for the now-enabled path.
- `yarn lint`, `yarn lint:i18n`, `tsc --noEmit` — clean.
- **On-device (iPad, real Guardian account, testnet):** deposited 5 USDC end-to-end. Collateral built as a Miden P2IDE, Guardian co-signed and submitted on-chain (txid `0x17a9999bebf8b6b9c73d696571361d2a04a20c875edb367b6a70cb69d851798c`), completion marked "Deposited to lending", the Epoch allocator confirmed (`epochStatus: confirmed`), and the position rendered in the Earn UI (Total Deposited $0 → $4.96).


0xnullifier (approved):
LGTM

PR #446, opened 2026-08-03
Order assets by fiat value

### LINKED ISSUE #432: Order assets on explore page via some criteria
The assets on the home page should be ordered by some criteria (I think fiat amount is fine)

0xnullifier:
## Summary

- keep the native asset pinned to the top of the Home asset list
- order all remaining assets by descending current USD value
- add regression coverage for native-first fiat ordering

## User impact

Users see their most valuable non-native holdings first while the native asset remains immediately accessible at the top.

## Testing

- `yarn test src/app/pages/Explore.test.tsx --runInBand` (23 tests passed)
- `yarn lint`
- `yarn ts`

Fixes #432


WiktorStarczewski (approved):
LGTM

PR #448, opened 2026-08-03
fix(android): adaptive launcher icon safe zone + drawer safe-area padding

### LINKED ISSUE #429: Android: bottom content clipped `env(safe-area-inset-bottom)` is 0 in WebView under Android 15+ enforced edge-to-edge
## Symptom

Bottom-of-screen content is clipped on Android — reported repeatedly in the feedback DB (tickets 104/105/107: General settings / Address book / Recovery on a Galaxy Tab S9 FE+ 5G; tickets 78/79 on a Galaxy A27). `mobile.html` applies

```css
padding-bottom: max(16px, env(safe-area-inset-bottom)) !important;
```

which handles this correctly on iOS, but on Android the padding resolves to just 16px and content ends up under the navigation/gesture bar.

## Why `env(safe-area-inset-bottom)` is 0 on Android

Two things compound:

1. **The app opts out of edge-to-edge, but Android 15+ ignores that.** `MainActivity.setupStatusBar()` calls `WindowCompat.setDecorFitsSystemWindows(window, true)` (`android/app/src/main/java/com/miden/wallet/MainActivity.java:32`). On Android ≤14 that means the system insets the window itself, so `env()` = 0 is *correct* and the 16px fallback is fine. But we build with `targetSdkVersion = 36` (`android/variables.gradle:4`), and starting with Android 15 (API 35) edge-to-edge is **enforced** for apps targeting 35+ — `setDecorFitsSystemWindows(true)` is deprecated/ignored, and with target 36 the opt-out (`windowOptOutEdgeToEdgeEnforcement`) is removed entirely. So on Android 15/16 devices the WebView is drawn under the status bar and nav/gesture bar whether we like it or not.

2. **Android WebView does not report nav-bar insets through `env(safe-area-inset-*)`.** Chromium's WebView only surfaces *display-cutout* insets via `env()` (and only with `viewport-fit=cover`); the navigation-bar inset is not included (WebView edge-to-edge inset support is still behind a rollout flag upstream). So even though the window is now edge-to-edge, the page still sees `env(safe-area-inset-bottom) = 0`.

Net result on Android 15+: content drawn under the system bar with only 16px of padding → the clipped buttons/text in the reports above. (Same mechanism affects the top: `window.setStatusBarColor(Color.WHITE)` is also deprecated/ignored under enforced edge-to-edge, which likely explains some status-bar-overlap/appearance oddities too.)

## Suggested fix

Don't rely on `env()` on Android. Feed real insets from the native side:

- In `MainActivity`, embrace edge-to-edge (`setDecorFitsSystemWindows(window, false)` on all versions for consistency) and attach `ViewCompat.setOnApplyWindowInsetsListener` on the WebView/decor view, forwarding `Insets` (systemBars + displayCutout + ime as appropriate) into the page as CSS custom properties (e.g. `--android-safe-area-bottom`) via `evaluateJavascript`.
- In `mobile.html`, change the body padding to `max(16px, env(safe-area-inset-bottom), var(--android-safe-area-bottom, 0px))` (and same idea for top/left/right).
- Alternatively adopt a maintained plugin that does exactly this (e.g. the community `@capacitor-community/safe-area` edge-to-edge support) rather than hand-rolling.

Feedback DB refs: tickets 78, 79, 104, 105, 107 (Android settings-screen clipping cluster).

0xnullifier:
Closes #429

## Drawer clipped under the Android navigation bar

Bottom-sheet drawers (e.g. Guardian Settings → "Rotate Guardian") were clipped by the system navigation bar on Android.

The diagnosis in #429 turned out to be wrong: `env(safe-area-inset-bottom)` **is** honored by the Android WebView — `mobile.html`'s body padding works fine, as regular page content is not clipped. The actual bug is that the shared `DrawerContent` (`src/lib/ui/drawer.tsx`) is `position: fixed; bottom: 0`, which anchors to the **viewport** — body padding never applies to fixed elements, so the sheet slid under the nav bar.

Fix: the drawer now pads itself with `env(safe-area-inset-bottom)` (resolves to 0 on extension, so no layout change there). `GuardianInfoDrawer`'s per-instance workaround is removed to avoid double padding.

## Launcher icon rendered as a cropped orange blob

The adaptive-icon foreground PNGs drew the logo across ~66% of the 108dp canvas — right at the edge of the launcher mask, so circular masks zoomed into the middle of the artwork, showing only orange with the white "B" counter-shape.

Fix: rescaled all `mipmap-*/ic_launcher_foreground.png` so the logo sits centered at ~44% of the canvas (a square-ish logo must fit the mask circle's inscribed square, not the full 66dp safe-zone width). Legacy `ic_launcher.png`/`ic_launcher_round.png` were already correct and are untouched.


WiktorStarczewski (approved):
LGTM

PR #489, opened 2026-08-04
test(e2e): deterministic guardian browser-crash recovery test + harden reopen() cold-SW readiness

WiktorStarczewski:
Follow-up to #450 (the crash-resilient `reopen()` that recovers the guardian recovery specs from an intermittent, non-OOM Chromium crash by relaunching the persistent context from the on-disk profile). That fix shipped but its recovery path had never actually been exercised on CI — both green runs happened to not crash. This PR deterministically verifies it and hardens it.

## What this adds

- **`killBrowser()` harness method** — tears down the whole browser (persistent context + service worker), not just the page, as a **deterministic** stand-in for the intermittent Chromium crash (where an incidental one is a coin flip). Verified locally against `playwright-core@1.61.0`: `context.browser()` is non-null for a persistent context, `.close()` flips `isConnected()` to false, so `reopen()` genuinely takes its relaunch branch.
- **Harden `reopen()`'s post-relaunch readiness** — the relaunched context has a **cold** SW that must re-load the ~14MB WASM before intercom answers. The old readiness wait was a single 90s window; the initial launch budgets 3×90s *with reloads* for exactly this. `reopen()` now does the same **only on the relaunch path** (the fast warm-swap path is byte-for-byte unchanged). `reopen()` returns whether it relaunched.
- **One honest deterministic test** — `killBrowser()` → `reopen()` → assert the wallet recovers to a usable state with its account intact (read back from the on-disk IndexedDB profile), plus a **tripwire** asserting `reopen()` actually took the relaunch path (so the test can't silently stop covering the crash path).

## What it deliberately does NOT claim

An earlier draft of this test asserted the interrupted rotation **resumes to completion** after a crash. An adversarial review caught that this is the **opposite** of production behavior and confirmed it against `main`:

- Production's cold-start handler (`runtime.onStartup` → `failInterruptedTransactions`, `src/background.ts` / `transaction/cancel.ts`) deliberately **fails** every in-flight row and requires a manual **Retry** — resubmitting a tx that already landed on-chain would trip the node's nullifier check.
- The unpacked test extension (`--load-extension`) never fires `runtime.onStartup` on relaunch (Chrome fires `runtime.onInstalled` instead), so the harness can't reproduce that production path anyway.

So the test asserts only the harness-level guarantee (browser dies → wallet comes back usable), not transaction resume. Production's crash behavior is **fail + retry** — escapable, not stuck.

## Follow-up (not in this PR)

A faithful "crash mid-rotation → rotation Failed → Retry completes" test needs a `MIDEN_E2E_TEST`-gated hook to fire the `onStartup` sweep in the harness. Separately worth investigating: the on-chain-committed-but-guardian-not-registered window (does Retry/sync fully re-register the new hot key with the guardian, or can it 401 — a real "stuck guardian" risk on a security-sensitive path).

Harness-only change (`playwright/e2e/**`); no wallet source or user-facing behavior changes.

### DIFF
--- a/playwright/e2e/helpers/wallet-page.ts
+++ b/playwright/e2e/helpers/wallet-page.ts
@@ -261,8 +261,16 @@ export interface ChromeWalletPageApi extends WalletPage, IdbDumpSource {
    * password if needed. Used to prove state that must be durably persisted
    * (e.g. a switched guardian endpoint) survives a fresh session, not just
    * the one that made the change.
+   *
+   * Returns whether it took the crash-recovery RELAUNCH path -- i.e. the browser
+   * was dead, so the whole persistent context was relaunched from the on-disk
+   * profile with a cold SW -- vs a normal warm page swap. A spec that used
+   * `killBrowser()` can assert the return is `true` to prove the crash path
+   * actually ran (guarding against e.g. a Playwright change that made
+   * `killBrowser()` a no-op, which would otherwise pass silently on the fast
+   * path).
    */
-  reopen(): Promise<void>;
+  reopen(): Promise<boolean>;
   /**
    * Terminate the wallet mid-flight: closes the current page and, unlike
    * `reopen()`, does NOT open a fresh one -- models a user closing the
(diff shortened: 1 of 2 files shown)


WiktorStarczewski:
## Follow-up: crash → Retry recovery investigation (and why the "faithful" E2E test was dropped)

I investigated the two follow-ups named in the PR description. Outcome:

### 1. Verdict: a browser crash mid-rotation RECOVERS on Retry (for a reachable guardian)

Traced the full path (`HotKeyRotationGate.onRetry` → `ensureRotationTx(adopt=false)` → `initiateReplaceHotKeyTransaction` → `createReplaceHotKeyProposal` → `completeReplaceHotKeyTransaction` → `reRegisterCurrentStateOnGuardian`):

- Retry mints a **fresh** hot key (`new-hot-2`) and submits a **new** `replace_signer` targeting `[new-hot-2, cold]`. On-chain is `[new-hot-1, cold]`, so the transition is non-degenerate and **lands** — it is *not* a nullifier collision (refuting the "retry can't land because on-chain already equals new-hot" hypothesis).
- The completion handler re-registers the guardian **cold-signed**. Because `cold` is in every allowlist the account has ever held, `/configure` authenticates against the stale allowlist and rewrites it to `[new-hot-2, cold]`. Final state (on-chain, guardian allowlist, vault) is consistent; new-hot-2 requests authorize. **Recovered.**

### 2. Latent risk found (real, worth a look — NOT the crash path)

`completeReplaceHotKeyTransaction`'s cold re-register is **best-effort**: its try/catch only *logs* on failure, then unconditionally `swapHotKey`s, **clears `requiresHotKeyRotation`, and dismisses the gate**. So if the guardian is unreachable across all 5 retries *at completion time*, you get on-chain `[new-hot-2, cold]` but a **stale guardian allowlist**, flag cleared, gate gone, no Retry button. The only remaining self-heal is AutoSync — which is **hot-signed** (`syncGuardianAccounts` binds a service on `hotPublicKey`), so it can't recover a diverged allowlist when the new hot key isn't yet allowlisted (raw 401 → `runSync` rethrows before Stage-2 re-register → evict/rebuild same hot service → permanent loop).

Asymmetry to flag: **completion-time re-register is cold-signed (robust); the sync-time self-heal is hot-signed (fragile).** The completion handler clearing the flag on a best-effort failure converts a still-recoverable gated state into a potentially unrecoverable one. Whether it's *actually* permanent depends on the Rust guardian server's `/configure` policy (does it authorize a request signed by a key in the new cosigner set but not the current allowlist?) — that's in `OpenZeppelin/guardian`, not this repo. If `/configure` is allowlist-gated (which the pervasive 401 handling and the deliberate cold-signing strongly imply), it's permanently stuck.

### Why I did NOT add a "faithful" crash→fail→retry E2E test

A test that force-fails the in-flight rotation (via the cold-start `failInterruptedTransactions` sweep) and asserts Retry recovers **cannot be written faithfully in this harness**: production orphans the row by *killing the SW* in the crash, but the harness's SW stays **alive** and completes the *original* rotation when its delayed `/configure` resolves — clearing the flag before the Retry's fresh rotation finishes. `completeHotKeyRotation()` would then pass via the **original** completing, not the Retry path — the same false-coverage trap flagged earlier in this PR. Controlling the sweep-before-resume ordering that production gets for free isn't reliably reproducible from a spec, so a "faithful" test here would be either flaky or falsely-green. The deterministic `killBrowser()` + `reopen()` test in this PR covers what the harness *can* prove honestly (the browser dies → the wallet relaunches from disk to a usable state); the recovery-to-completion question is answered by the trace above.

Issue #473, opened 2026-08-04
Android delegated proving times out when the transaction pane is hidden


BrianSeong99:
From feedback ticket #139 (https://app.notion.com/3ae99411cf9081888cc0f331767dc167). Reported on Android.

## Symptom

Hiding the transaction pane or switching to another app caused repeated remote-prover timeouts. The retries later succeeded. When the pane remained visible, proving was reliable.

## Expected

Delegated proving and transaction tracking should continue when the app is backgrounded or the progress pane is hidden. UI lifecycle changes should not cancel, suspend, or starve the request.

## Investigation needed

Compare foreground and background behavior for the native bridge, WebView lifecycle, network access, timers, and transaction queue. Confirm whether the timeout occurs at the remote prover or because the app stops processing its response.

## Related

#419 handles general remote-prover unavailability and fallback. This issue is the Android lifecycle-dependent timeout trigger.

Feedback DB ref: ticket 139.

WiktorStarczewski:
Opened #658 for the false-**cancel** half of this.

Root cause (shared TS): on Android the WebView main-thread JS is frozen while backgrounded, but `Date.now()` keeps advancing, so the stuck-transaction reaper's wall-clock check (2 min on mobile) reaps a delegated prove that merely sat through a background stretch — surfacing as "Transaction took too long to process and was cancelled" on resume. The PR counts only active (foreground) time toward the threshold (tracks `document.hidden` intervals and subtracts them; mobile-only, desktop unchanged).

Honest scope: this does **not** touch the SDK's own ~10s remote-prover RPC deadline, which is the more likely source of the literal *"repeated remote-prover timeouts that retries later succeeded"* — a separate wall-clock timer that also elapses during a freeze (`resolveTransactionErrorMessage` only emits the "proving service timed out" copy on a `/timeout/i` match, which the reaper's message isn't). Fully closing this likely also needs a larger/background-aware remote-prover timeout on mobile (the ~10s default is already documented as "too tight" in `transaction/index.ts`) and/or a native keep-alive so the WebView isn't frozen mid-prove. Both need on-device verification — left as follow-ups. Couldn't repro end-to-end on the iOS sim here, so the reaper's contribution should be confirmed on a device.

PR #498, opened 2026-08-05
feat: improve native wallet funding flow

### LINKED ISSUE #425: Make the fund now prompt more robust
Change the fund now process into a drawer where user can see error messages as well if the faucet fails

### LINKED ISSUE #471: Unexpected IMIDEN appears after funding or minting
From feedback ticket #131 (https://app.notion.com/3ae99411cf9081b783ede625ebf4320c), with related context in ticket #89. Reported on iOS and Chrome.

## Problem

After minting, the tester received 100 MIDEN and an unexpected additional 10 IMIDEN. The source and purpose of the second asset were not explained.

## Expected

The funding/minting flow should provide exactly the documented assets. If IMIDEN is intentionally included, the UI should state why and how it differs from MIDEN; if not, the extra asset should not be created or surfaced.

## Validation

Trace the faucet/mint response, created notes, asset metadata, and final balance to determine whether this is intended test-token behavior, a routing defect, or a display/metadata defect.

Feedback DB refs: tickets 131 and 89.

0xnullifier:
## Summary

- Stop Fund Now from also requesting the legacy iMiden test asset from Forkchoice, so funding requests only native MIDEN.
- Open Fund Now in a TabLayout-owned drawer with automatic request start, loading and success feedback, detailed faucet errors, retry, explicit close, and mobile back handling.
- Keep the request lifecycle outside the Home component so closing the drawer or remounting the tab layout does not cancel, duplicate, or lose the in-flight result.
- Put the network-discovered native MIDEN asset first in the swap token list while retaining TOKEN_IMIDEN as an available swap asset.

## User impact

Funding an empty wallet now produces only the documented native MIDEN asset. Users can follow faucet progress in a dedicated drawer, safely close it while the request continues, and retry from the displayed error if the faucet fails.

## Testing

- 224 focused Jest tests across wallet funding, prompts, TabLayout, swap flow, and transaction summaries
- yarn ts --pretty false
- Prettier check for all changed TypeScript and locale files
- New funding drawer and lifecycle files: 96.61% branch coverage combined

Closes #471
Closes #425


0xnullifier:
closes https://github.com/0xMiden/wallet/issues/473

WiktorStarczewski (changes requested):
**`src/lib/wallet-funding.ts` — funding can report false success for the wrong account**

`openWalletFunding(address)` drops the new `address` whenever a request is already in flight: it takes the early `publish({ ...walletFundingState, open: true })` branch and keeps the previously-requested account's address. The drawer's mount effect then calls `startWalletFunding()`, which returns the existing shared promise (`if (fundingPromise) return fundingPromise`), so the reused request funds the **old** account while the UI reports *its* success for the newly-selected one — and `completeWalletPrompt(WalletPromptType.Faucet)` is persisted.

**Repro:** empty account A → Fund Now (faucet PoW runs, several seconds) → close the drawer (a supported action) → switch to empty account B → Fund Now. `openWalletFunding('B')` sees `fundingPromise` set, keeps `address = 'A'`; the drawer shows loading then "MIDEN requested successfully" and the Faucet prompt is marked complete — but **B received nothing**.

This is currently locked in by the new test (`wallet-funding.test.ts`, the *"continues after the drawer and its subscriber unmount while sharing one request"* case), which asserts the address stays `account-a` after `openWalletFunding('account-b')`.

**Suggested fix:** when a request is in flight, only reuse it if it targets the same address — if `fundingPromise` is set and `walletFundingState.address !== address`, treat it as a new target (queue it to start once the current request settles, or surface that a different account's request is still running) rather than silently re-showing the old request's outcome for the new account.

0xnullifier:
Superseeded by #504

PR #500, opened 2026-08-05
fix(android): keep the keyboard as an overlay so inputs stay visible

### LINKED ISSUE #496: Android: Keyboard pushes Send and Swap input fields behind the page header
## Problem

On Android, tapping certain input fields in the Send and Swap flows causes the page content to shift upward behind the fixed header when the numeric or text keyboard opens.

The affected fields include:

- The wallet address input field on the Send page.
- The amount input field on the Send page.
- The amount input fields on the Swap page.

After the keyboard opens, the focused field and surrounding content become partially clipped or hidden behind the page header. A large empty area may also appear between the visible content and the keyboard.

### Send — wallet address field

1. Open the wallet on Android.
2. Tap **Send**.
3. Tap the wallet address input field.
4. Observe the layout after the keyboard opens.

### Send — amount field

1. Continue to the amount-entry step in the Send flow.
2. Tap the amount input field.
3. Observe the layout after the numeric keyboard opens.

### Swap — amount field

1. Open the **Swap** page.
2. Tap one of the amount input fields.
3. Observe the layout after the numeric keyboard opens.

## Actual behavior

- The focused input field is pushed upward behind the fixed page header.
- The input text may become partially visible or completely hidden.
- Other page controls may also become clipped.
- A large unused blank area can appear above the keyboard.
- The user cannot clearly see or verify the wallet address or amount being entered.

## Expected behavior

When the keyboard opens:

- The focused input field should remain completely visible.
- The page content should remain below the fixed header.
- The layout should resize or scroll correctly based on the available viewport.
- No excessive blank space should appear.
- The user should be able to review the wallet address or amount before continuing.

## Environment

- Platform: Android
- Device: Samsung Galaxy A27
- Keyboard: Samsung keyboard
- Theme: Dark mode
- Wallet version: Not provided

## Regression context

This may be related to or a regression of #415, which addressed mobile keyboard resizing and input fields being hidden behind the keyboard.

The current behavior suggests that keyboard handling is still inconsistent across shared Send and Swap layouts.

Image

0xnullifier:
## Problem

On Android, the soft keyboard pushes the Send and Swap input fields behind the page header. A large empty area shows above the keyboard. Issue #496 reports this defect.

## Cause

The configuration `Keyboard.resize: 'none'` in `capacitor.config.ts` operates only on iOS. The Android Keyboard plugin ignores this configuration. On Android, two mechanisms decrease the layout at the same time:

- The operating system decreases or moves the window.
- The CSS inset `--keyboard-height` decreases the layout again.

The layout loses two times the keyboard height. The content moves behind the header.

## Correction

- Set `android:windowSoftInputMode="adjustNothing"` on `MainActivity` in `AndroidManifest.xml`. With this value, Android does not change the window when the keyboard opens.
- The CSS inset is now the only compensation. Android and iOS obey the same overlay contract.
- Add a comment in `capacitor.config.ts`. The comment tells you to keep the two configurations in agreement.

## Test

- We made a test build on the Android emulator (Pixel 10) and on a device.
- Open the Send page. Touch the address field. The field stays visible above the keyboard.
- Do the same test on the amount field and on the Swap amount fields. The layout does not move behind the header.

Closes #496

### DIFF
--- a/capacitor.config.ts
+++ b/capacitor.config.ts
@@ -23,7 +23,14 @@ const config: CapacitorConfig = {
       // No special config needed
     },
     Keyboard: {
-      // Prevent keyboard from pushing content - overlay instead
+      // Prevent keyboard from pushing content - overlay instead.
+      // iOS-ONLY: the Android Keyboard plugin ignores `resize` entirely
+      // (it only reads `resizeOnFullScreen`). Android instead pins
+      // android:windowSoftInputMode="adjustResize" on MainActivity so the
+      // native window resize is the single compensator there; the JS
+      // --keyboard-height inset (lib/mobile/keyboard-inset.ts) is gated to
+      // iOS only. Changing either side alone double-compensates or removes
+      // all compensation — keep the pair in sync.
       resize: 'none'
       // The iOS accessory bar (Done + arrows) — the only way to dismiss the
       // number pad, which has no return key — is enabled at RUNTIME via


0xnullifier:
Superseded by #499, which fixed the same double-inset bug (released in v1.15.19) by pinning adjustResize on Android and gating the --keyboard-height inset to iOS. Closing.

PR #501, opened 2026-08-05
fix(mobile): hot-key plugin fixes and backwards compatibility

### LINKED ISSUE #482: Android Fund Wallet claim fails when secure hardware is unavailable
From feedback ticket #157 (https://app.notion.com/3b199411cf9081f6a951fcc701199ccd). Reported on a Galaxy S23 running Android 16 with the LambdaClass Guardian.

## Problem

Claiming the note created by **Fund my wallet** fails with:

`[HARDWARE_UNAVAILABLE] Secure hardware unavailable: null`

The carousel also shows the raw hardware-unavailable error, and the user cannot complete the funding claim.

## Expected

- Detect secure-hardware capability before starting the claim.
- Use the supported fallback when one exists, or explain the device/profile requirement before the transaction begins.
- Never surface a `null` detail as the user-facing explanation.
- Preserve the note in a safely retriable state.

Feedback DB ref: ticket 157.

0xnullifier:
## Android (`HotKeyPlugin.kt`)

**Release blockers from the audit**
- **OAEP portability**: AndroidKeyStore only reliably executes OAEP with MGF1-SHA-1 (API ≤ 33 rejects other MGF1 digests at `Cipher.init`; API ≥ 35 enforces the `RSA_OAEP_MGF_DIGEST` authorization, SHA-1-only for keys minted without the tag). New blobs are always wrapped with MGF1-SHA-1 (no security loss — OAEP does not rely on MGF1 collision resistance); unwrap tries SHA-1 then SHA-256 so pre-existing blobs keep working. Previously, generate succeeded (software-provider encrypt) while every sign failed on Android ≤ 13 and on some secure elements.
- **Proven generation**: `wrapSecretProven` — generate wrapper key → verify `KeyInfo` is StrongBox/TEE-backed (software-backed rejects as `HARDWARE_UNAVAILABLE` → JS fallback) → wrap → **round-trip decrypt probe**. StrongBox attempt first with a full ordinary-TEE retry if *any* step fails (field case: S23 eSE accepted key generation but failed the first real OAEP unwrap). A present-but-failing StrongBox is surfaced to the user via the existing hardware-report prompt (`strongBoxError` → `reportHotKeyHardwareFailure`).
- **Legacy biometric path (Android 9–10)**: SDK-gated authenticators (`BIOMETRIC_STRONG|DEVICE_CREDENTIAL` is invalid pre-API-30 and threw synchronously, permanently wedging the plugin into `BIOMETRIC_BUSY`), negative button pre-30, try/catch around prompt build/authenticate, pending state consolidated into one object with a stale-callback guard and `handleOnDestroy` cleanup.

**Hardening**
- `revealHotKey` requires fresh native user presence (BiometricPrompt, biometric or device credential) before returning the scalar; rejects `AUTH_UNAVAILABLE` when the device has no usable authenticator instead of skipping the gate.
- Stable error codes: `UNWRAP_FAILED` (blob/key mismatch → rotate, not "defective hardware"), `KEY_INVALIDATED`, `BAD_CIPHERTEXT`, `INVALID_INPUT`, `KEY_NOT_FOUND`; raw provider messages sanitized (no more `Secure hardware unavailable: null`).
- Strict bounded wire-format parsing before any decode; strict `deleteHotKey` (missing alias resolves, real Keystore failure rejects); scalar rejection-sampled into `[1, n)` and re-validated at sign; all four ECDSA recovery-id candidates; honest "hardware-wrapped at rest" docs; prompt title fixed ("Bread" → "Miden Wallet").

## iOS (`HotKeyPlugin.swift`) parity
- Presence-gated reveal via `LAContext` device-owner auth (Face ID / Touch ID / passcode), `AUTH_UNAVAILABLE` when none available.
- Generate-time ECIES round-trip probe; scalar rejection-sampling; strict bounded ciphertext parsing (`BAD_CIPHERTEXT` / `INVALID_INPUT`); strict delete (`errSecItemNotFound` = success, other failures reject); security-model docs aligned with Android.

## JS
- `generateHotKey` result carries optional `strongBoxError`; the facade surfaces it through the existing `HotKeyHardwareUnavailable` prompt (copyable raw error) without failing generation. Prompt body copy neutralized to cover both degraded-to-TEE and full-fallback cases.

## Notes
- Existing wallets whose blobs are already un-decryptable (Android ≤ 13, or MGF-incompatible secure elements) need a hot-key rotation; they now get an actionable error instead of a hardware report. Verified in the field: an S23/Android 16 wallet that previously failed every sign works after rotation, StrongBox-backed.
- Android: `compileDebugKotlin` clean, 27/27 secure-hot-key jest tests pass. iOS changes still need a device/simulator build pass.
- Follow-ups (not in this PR): map `UNWRAP_FAILED`/`KEY_INVALIDATED` to a rotation prompt in the JS facade; guardian-side policy for spend-domain hot-key signatures.

closes #482


WiktorStarczewski (commented):
Minor — a small Android regression in the reveal path (not blocking):

`confirmPresenceThenReveal` gates the (CryptoObject-less) presence check on `BiometricManager.canAuthenticate(allowedAuthenticators())`, and `allowedAuthenticators()` returns `BIOMETRIC_STRONG` only on API < 30. On a device with a secure lock screen (PIN/pattern) but **no enrolled strong biometric**, `canAuthenticate` returns `NONE_ENROLLED`, so Reveal Hot Key is rejected with `AUTH_UNAVAILABLE` — whereas it resolved silently before this PR.

Since this prompt carries no `CryptoObject`, a device-credential fallback is safe here — consider allowing `DEVICE_CREDENTIAL` for the `cipher == null` presence path (at least on API 29+) rather than reusing the CryptoObject-constrained helper. Narrow impact (users with a biometric sensor can enroll and retry; signing is unaffected — only the reveal/backup flow), hence minor.

WiktorStarczewski (changes requested):
Thanks — this is a follow-up from a deeper multi-pass review. The StrongBox/TEE hardening and key-wrapping work is valuable, but there's one user-facing correctness issue I'd like fixed before merge.

### Requesting changes

**Normal Cancel / no-enrolled-auth is misreported to the user as a "secure hardware failure."**

The new mandatory presence gate on `revealHotKey` (iOS `LAContext.evaluatePolicy(.deviceOwnerAuthentication)`; Android `confirmPresenceThenReveal`) rejects with `USER_CANCELLED` / `AUTH_UNAVAILABLE` / `BIOMETRIC_BUSY`. But `withHardwareFailureReport` (`nativePlugin.ts:63-68`) catches *all* of these and calls `reportHotKeyHardwareFailure(...)`, which persists state and seeds `WalletPromptType.HotKeyHardwareUnavailable`.

Repro: Settings → Reveal Hot Key → the Face ID / biometric sheet appears → user taps **Cancel** (or the device has no biometric/passcode enrolled). The user is then shown a "secure hardware failure" and the hardware-unavailable prompt is seeded — for what is a completely normal cancel/no-auth outcome.

Please filter `withHardwareFailureReport` so it only reports genuinely actionable hardware codes (e.g. `ERR_HARDWARE_UNAVAILABLE`, `strongBoxError`), and treat `USER_CANCELLED` / `AUTH_UNAVAILABLE` / `BIOMETRIC_BUSY` as ordinary non-fatal outcomes (and route `UNWRAP_FAILED` / `KEY_INVALIDATED` to a rotation-prompt UX rather than "hardware failure").

### Related — please confirm intent

- **Reveal Hot Key becomes unusable for users with no OS-level lock.** The gate now requires an OS authenticator (`canAuthenticate(BIOMETRIC_STRONG|DEVICE_CREDENTIAL)` / `canEvaluatePolicy(.deviceOwnerAuthentication)`). A Guardian-hot-key user who protects the app only with the wallet's own PIN and has never enrolled a device lock — a supported config — loses a Reveal that worked before. Either confirm this trade-off is intentional (and note it in release/support docs) or fall back to the app-level unlock when no OS authenticator is enrolled.

### Minor / non-blocking (test coverage)

- The new `strongBoxError` branch in `generateHotKey()` (`nativePlugin.ts:31-41`) has no test. Add cases: reports on `strongBoxError` present; still resolves when `reportHotKeyHardwareFailure` itself rejects.
- The ~500 lines of new Kotlin key-wrapping/retry/fallback logic (`HotKeyPlugin.kt`) ship with no automated tests. Extracting the pure-logic pieces (`parseCiphertext` bounds, `rejectClassified` dispatch, the rejection-sampling predicate) into plain-JUnit-testable functions — no Android Keystore needed — would cover the both-attempts-fail and legacy-blob (MGF1 SHA-1 → SHA-256) paths this PR was written to fix.

PR #502, opened 2026-08-05
fix(private-send): relay note before waiting for commit so the block hint stays below the commitment block

WiktorStarczewski:
## Problem

Private-note delivery over the note-transport layer can silently fail: the recipient never receives the note.

In `completeCustomTransaction` (and the guardian/delivering path), the wallet relays the note **after** `waitForTransactionCommit`:

```ts
await midenClient.waitForTransactionCommit(...);   // sync height advances to/past the note's commitment block C
await midenClient.sendPrivateNote(fullNote, ...);  // relays block hint = get_sync_height()  → ≥ C
```

`sendPrivate` takes no hint parameter, so the SDK always relays the client's **current sync height** as the note's `after_block_num`. The recipient imports the note as an `ExpectedNote` and scans **forward** from that hint for the on-chain commitment. Because we wait for commit first, the hint is at or above the commitment block `C` — so on any chain that produces a block between commit and relay, `hint > C`, the recipient scans past the commitment and **never finds the note**.

This is latent on testnet (slow blocks → sync height stays ≈ `C` at relay), which is why it hasn't surfaced. It reproduces deterministically on a fast devnet.

## Fix

Relay **before** waiting for commit — the SDK's documented "prompt relay" flow — so the hint is captured while sync height is still below `C`. The recipient then scans forward from below the commitment and picks it up when it lands. The commit wait stays (after the relay) so `Completed` status is still gated on on-chain confirmation.

Applied at both private-relay sites in `complete.ts`.

## Verification

Reproduced and fixed on a fast custom devnet via the blockchain e2e harness:
- Before: `send-private` timed out at 5m every run (wallet B balance stayed 0).
- After: `send-private` passes in ~1.4m (B receives); `send-public` still passes (no regression).

## Follow-up (SDK)

The more robust fix belongs in the SDK: `send_private_note` should relay the note's **actual commitment block** (from its inclusion proof) as the hint rather than the sync height — exactly what the `miden-client` CLI's `notes --send` already does. That would let the wallet keep waiting for commit before relaying. Filing separately.

### DIFF
--- a/.github/workflows/pr-e2e-local.yml
+++ b/.github/workflows/pr-e2e-local.yml
@@ -8,7 +8,25 @@ permissions:
   contents: read
 jobs:
   chrome-local:
-    name: local-e2e (chrome)
+    # Run the core suite under BOTH block cadences: the miden-node default (3s
+    # block / 1s batch, realistic timing) and fast 500ms blocks. Fast blocks
+    # guarantee a block lands in the wallet's commit->relay window, making
+    # send-private a deterministic guard for the private-note block-hint overshoot
+    # (web-sdk#262 / wallet#502). The fast leg is scoped to that one timing-critical
+    # spec to stay quick and stable; the default leg keeps full-suite coverage.
+    # The default leg's name stays exactly "local-e2e (chrome)" so any required-check
+    # mapping is unaffected.
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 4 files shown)


0xnullifier (approved):
Lgtm

Issue #476, opened 2026-08-04
Fund Wallet needs a clear pending state and less alarming recovery copy


BrianSeong99:
Consolidates feedback ticket #144 (https://app.notion.com/3b199411cf908138a1a9c67266479804) and the funding/recovery context in ticket #159 (https://app.notion.com/3b199411cf9081aca177d28c6f3963c0).

## Problem

On iOS, roughly 30 seconds passed after **Fund this wallet** before the note/free-float notification appeared, with the network transaction completing within about a minute. During that gap, the flow did not make the pending state clear.

The recovery/security prompt then appeared immediately after funding, making the tester think recovery confirmation was required to receive the free tokens. Copy about “bad actors” also felt more alarming than informative.

## Expected

- Show an explicit funding state from request through note availability and consumption.
- Separate funding completion from the recovery-confirmation prompt so they do not appear causally linked.
- Use neutral, concrete security copy that explains the benefit without implying an immediate threat.

Feedback DB refs: tickets 144 and 159.

BrianSeong99:
Have @dotsolo to implement an ux flow and then have @0xnullifier to implement the flow once its ready

PR #504, opened 2026-08-06
Fund-wallet flow: full funding/success lifecycle + prompt card redesign

dotsolo:
Partially addresses #476 (the "show an explicit funding state from request through note availability" part — the recovery-prompt separation and security-copy items from that issue are not covered here).

## Problem

On iOS, tapping **Fund this wallet** acked the faucet request within seconds, but the minted note only becomes visible after chain inclusion + a client sync (~30–60s). The prompt card flashed success on the ack and vanished, so the Home screen looked completely idle for the whole gap until the note notification appeared.

## What changed

**Funding lifecycle (HomePrompts):**
- The faucet card itself is now the tap target (no separate CTA button).
- Tapping enters a **"Funding" hero**: animated hourglass tile, indeterminate progress bar along the card's bottom edge, and *"Your funds will be available shortly"*.
- When the funds land (claimable note or balance), a green **"Funds deposited — You have $X ready"** success beat plays (~2.4s), then the prompt completes and hands off to the pending-notes card.
- The pending-notes card **holds back while the hero is on stage** — it sorts first in the carousel and previously would have pushed the success state off-screen the instant the note arrived.
- The wait **survives remounts and app restarts** via a persisted, per-account request marker (`faucet_funding_v2:<account>`), and falls back to the actionable card 3 minutes after the original request if the mint never lands.
- The pending-notes card's "See Now" button was folded into a whole-card tap as well.

**PromptCard:**
- New optional icon slot (coins icon on the faucet card), filled pill CTA, corner dismiss X that hides while a request is in flight, press-in feedback (`active:scale-[0.98]`) on tappable cards, and a reusable `hero` takeover mode (icon tile + wordmark + sub-line lockup).
- New `hourglass` icon added to the v2 icon set.

**Polish:**
- Home "Assets" heading now uses the heading font (Nunito) — it was falling back to Inter.
- Onboarding confirmation screen: bold Nunito heading, Nunito body copy, and the daily-reminder note styled as a pill (also fixes a stray `font-medium` JSX attribute that was never applied).

## Rebase note

This PR now targets `main` directly (it previously stacked on `fix/remove-imiden-swap-faucet`, so until #498 merges the diff here includes that branch’s commits). Where the two collided (`HomePrompts`, its tests, locales), this branch's in-card funding lifecycle supersedes the base's `WalletFundingDrawer` wiring — the drawer and `lib/wallet-funding` remain in the tree but are no longer opened from the faucet card. The base's native-MIDEN-only `faucet()` and swap-token ordering are preserved and compose with this flow.

## Review fixes (round 2)

Addressed from internal review:

- **Per-account funding wait (critical).** The marker is stored per account and the on-screen wait is tagged with the requesting account; display, arrival, and backstop all refuse a wait that isn't the current account's. An account switch clears the presentation but leaves the other account's marker intact, so switching back (or restarting) resumes the right account's wait — and a request that acks or fails after switching away can't paint the new account's card.
- **Baseline-gated arrival.** The marker carries the note ids that existed at request time; only a note beyond that baseline (or a balance) counts as the mint landing, so a pre-existing claimable note can't fake an instant success.
- **Request-anchored backstop.** The 3-minute fallback is computed from `requestedAt`, not from the mount, so a resumed wait still ends 3 minutes after the original request.
- **60s faucet timeout.** `faucet()` rejects if the PoW/mint request hangs, so the loading hero can't hold (and suppress the pending-notes card) with no in-session recovery.

## Review fixes (round 3)

- **The 60s faucet timeout now cancels the underlying work.** `faucet()` aborts an `AbortController` when the timer fires; the signal is threaded through `getPowChallenge`/`requestTokens` (into `fetch`) and checked on every PoW iteration, so a delayed response can't mint a second note behind a retry and the PoW search stops burning CPU after the UI reports failure. Tests assert the signal actually aborts (and that the PoW loop stops), not just that the wrapper rejects.
- **Whole-card actions are keyboard-accessible.** The tappable `PromptCard` is now `tabIndex={0}` and activates on Enter/Space, with keys on the inner dismiss/CTA buttons ignored so one press can't fire both actions.
- **The 3-minute backstop is anchored to the tap.** `requestedAt` is captured before `await faucet(...)`, so a slow ack no longer stretches the wait toward ~4 minutes.

## Testing

- 101 tests pass across the funding-flow suites (`HomePrompts`, `wallet-prompts`, `wallet-funding`, `WalletFundingDrawer`, `TabLayout`), including coverage for: card-tap funding, the Funding→Funded! sequence, arrival via notes and via balance, the persisted-marker resume path, pending-notes suppression during the success beat, tap-to-retry after failure, the account-switch leak (hero stays off the other account, marker preserved, switch-back resumes), pre-existing-note no-instant-success, backstop timing anchored to the request (fake timers), per-account marker roundtrip/shape validation, and the faucet timeout.
- Verified end-to-end on the iPhone 17 simulator against testnet: onboard → tap card → Funding hero through the ~30s gap → Funds deposited beat → pending-notes hand-off.


BrianSeong99:
can you make your claude to update the description to include the original issue to this PR? @dotsolo

WiktorStarczewski (changes requested):
Really nice feature — the persisted marker for remount/restart survival is the right call, and a couple of genuinely subtle interactions are handled correctly: suppressing the pending-notes card while the hero is on stage (it sorts first and would otherwise shove the hero off-screen the instant the note lands), and ceding completion authority from the seeding effect to the hero sequence (the `!awaitingFaucetFunds && !faucetFundsArrived` guard). The comments explain *why* rather than *what*, which made this reviewable.

That said, there are three real correctness bugs — one critical — plus a couple of suggestions.

## 🔴 Critical — the funding marker is global; the "Funding" hero leaks across accounts

`FAUCET_FUNDING_REQUESTED_AT_KEY = 'faucet_funding_requested_at_v1'` (`src/lib/wallet-prompts.ts`) has no per-account scoping, and `awaitingFaucetFunds` / `faucetFundsArrived` live in `HomePrompts` state that is **not** reset on an account switch. `Explore` → `HomePrompts` gets `account` from `useAccount()` (Zustand `currentAccount`) and re-renders **in place** — there's no `key={account.publicKey}`, so switching accounts does not remount.

**Repro (both accounts unfunded):**
1. On account A, tap Fund → `awaitingFaucetFunds = true`, global marker stamped, Funding hero shows.
2. Switch to account B → `Explore`/`HomePrompts` re-render in place with B's balances/notes; nothing resets the flags → **B now shows the "Funding" hero** for A's request.
3. B is genuinely unfunded, so the arrival effect never fires; the spurious hero holds until B's backstop — and that backstop calls `setFaucetFundingRequestedAt(null)`, **erasing account A's still-in-flight marker**, so an app restart during A's real wait no longer resumes.

The app-restart variant is the same: a fresh mount on B reads the global marker and resumes A's hero on B.

**Fix:** suffix the storage key with `account.publicKey` (e.g. `faucet_funding_requested_at_v1:<pk>`), read/write it per-account, and reset `awaitingFaucetFunds` / `faucetFundsArrived` in an effect keyed on `account.publicKey`.

## 🟠 "Funded!" fires on *any* existing or unrelated note/balance — an instant false success

The arrival effect gates on `hasPendingNotes || hasBalance`, where `hasPendingNotes` is the raw count of **all** current claimable notes, not a delta from the request baseline. The faucet card is seeded purely on `!hasBalance`, and incoming Miden notes sit unconsumed (`hasBalance === false`) — so a user with a **pre-existing claimable note** and zero balance sees the faucet card.

**Repro:** such a user taps Fund → `fundWallet` resolves and sets `awaitingFaucetFunds = true`; the arrival effect sees `hasPendingNotes === true` on the very next commit and fires immediately → the Funding hero flashes for ~0 ms and the card jumps straight to **"Funds deposited — you have $X ready"** and completes the faucet prompt, while the actual faucet mint is still 30–60 s away. It also mis-attributes the pre-existing note's USD amount to the faucet. The milder variant (an unrelated deposit arriving mid-wait) has the same root cause.

**Fix:** snapshot a baseline at request time (a note-id set, or note count / balance) and only trip arrival when it *increases* beyond that baseline.

## 🟠 The 3-minute backstop is anchored to mount, not to `requestedAt` — resumed waits can run ~6 min

The backstop effect schedules `FAUCET_FUNDS_ARRIVAL_TIMEOUT_MS` (3 min) from when `awaitingFaucetFunds` flips true. On resume, the resume effect sets `awaitingFaucetFunds = true` for any marker younger than 3 min — so a marker that's 2:59 old starts a *fresh* 3-min backstop, i.e. ~5:59 of "Funding" total if the mint silently failed. It's bounded, but the value is wrong.

**Fix:** schedule the backstop for `max(0, requestedAt + FAUCET_FUNDS_ARRIVAL_TIMEOUT_MS − Date.now())`, reading `requestedAt` from the marker.

## 🟡 Suggestions

- **`faucet()` has no timeout, and the initial `loading` hero has no backstop.** `faucetHeroActive` includes `faucetStatusIndicator === 'loading'`, but the backstop only covers `awaitingFaucetFunds`, and `faucet()` is a bare `fetch` with no `AbortController`. If the request hangs, `fundingRef` stays `true`, the loading hero holds, and the pending-notes card stays suppressed with no in-session recovery (no marker was stamped yet, so it only self-heals on restart). Consider an `AbortController` timeout on the faucet call.
- **Test gaps** let the three bugs above ship: no account-switch test (critical #1); the tap test starts with `claimableNotes={[]}` and only adds notes on rerender, so the pre-existing-note instant-false-success (#2) is never exercised; and there's no backstop-timing assertion (#3). The beat tests also use real timers with `waitFor(…, { timeout: 3500 })` — slow and flaky; prefer fake timers.
- Minor / not reachable today: the resume effect returns without clearing a stale marker if `faucetIsTerminal` is already true (cosmetic, overwritten next funding); the non-dismiss/non-hero branch always renders a `ChevronRight` even when `onClick` is undefined (only matters if a non-dismissible prompt is ever added — all current ones are dismissible).

## ✅ Verified fine

`handleDismiss`/`handleAction` both `stopPropagation()`, so nested-button taps don't bubble to the card's `onClick`; the PromptCard hero/icon/dismiss/status branches all render correctly (an in-flight card correctly can't be dismissed); no timer leaks (beat + backstop each `clearTimeout` in cleanup; the removed `successTimerRef` is fully replaced); the duplicate `completePrompt(Faucet)` path is idempotent; i18n is fully wired across all locales including the `$amount$` placeholder, and both hero icons resolve; and `Confirmation.tsx` correctly removes an invalid bare `font-medium` JSX attribute (it had been emitting `font-medium="true"` on the DOM).

Happy to re-review once the account-scoping (#1) and the arrival-baseline (#2) are addressed.

dotsolo:
@WiktorStarczewski Just updated the PR and it should include the requested fixes and @BrianSeong99 it should also be updated to address the original issue!

BrianSeong99:
Reviewed the current head (`afee1949`). The persisted marker, note baseline, and request-relative resume backstop are solid improvements. I found two P2 issues and one lower-priority suggestion; no P0/P1 correctness issues.

## 🟠 P2 — the timeout does not cancel the underlying faucet work

`faucet()` uses `Promise.race`, but `getPowChallenge`, the PoW loop, and `requestTokens` receive no `AbortSignal`. Once the wrapper rejects after 60 seconds, the original work can continue in the background while the UI permits another attempt.

In the delayed-response case, both attempts could reach the faucet and mint notes. In the PoW case, the original loop can continue consuming resources after the UI reports failure.

**Fix:** propagate an `AbortSignal` through both fetch calls and check it inside the PoW loop. The test should verify cancellation, not only that the wrapper promise rejects.

## 🟠 P2 — whole-card actions are not keyboard accessible

`PromptCard` renders tappable cards as a `div role="button"` with `onClick`, but without `tabIndex` or Enter/Space handling. This PR moves the faucet and pending-notes actions from real buttons onto that card, making those actions unavailable to keyboard users.

**Fix:** add complete keyboard-button behavior or retain a semantic button for the primary action.

## 🟡 Suggestion — the backstop starts after the faucet acknowledgement

`requestedAt` is created after `await faucet(address)`, so "three minutes after the original request" can be up to roughly four minutes after the user taps when the faucet call approaches its 60-second timeout.

This remains bounded, so I do not consider it a blocker. Either capture the timestamp before awaiting the faucet or adjust the description to say the three-minute wait starts after acknowledgement.

## ✅ Verified fine

The persisted marker survives remounts, the baseline prevents a pre-existing note from causing immediate false success, and the resumed backstop is correctly calculated from its stored timestamp. The pending-notes suppression during the success beat also remains correct.

## PR housekeeping

- This PR is based on `fix/remove-imiden-swap-faucet` / #498 rather than `main`. Please retarget/rebase it onto `main`; stacked branches risk the child PR being closed when the base branch is merged and deleted.
- This implements only the funding-state portion of #476. "Partially addresses #476" would be more precise.
- Please remove the attribution footer from the PR body.
- GitHub currently reports no checks for this branch, so the locally reported tests are the only available verification.

src/main.css
```diff
@@ -528,3 +528,57 @@ body[data-hide-navbar]::after {
...
+  }
+  to {
+    transform: scale(1);
+    opacity: 1;
+  }
+}
+
+.prompt-hero-pop {
+  animation: prompt-hero-pop 240ms cubic-bezier(0.23, 1, 0.32, 1) both;
+}
```
0xnullifier, line 608: Is there a reason why you won't use framer motion here, I mean this works but we can do this by not polluting the main.css file when we already have framer motion and tailwind animate in the bundle

0xnullifier (changes requested):
Just a nit on why did we added animation to main.css when we can use framer or tailwind

dotsolo:
Addressed the review in 277872e:

- **P2 — timeout now cancels the underlying faucet work.** `faucet()` aborts an `AbortController` when the 60s timer fires; the signal goes through `getPowChallenge` and `requestTokens` into their `fetch` calls and is checked on every `solvePowChallenge` iteration. The `Promise.race` remains only as a guarantee the wrapper rejects on time even if the abort isn't observed promptly. Tests now assert cancellation itself: the signal handed to `mintFromMidenFaucet` flips to aborted on timeout, both fetches receive the signal, and an abort ends an otherwise-unsolvable PoW search.
- **P2 — whole-card actions are keyboard-accessible.** Tappable `PromptCard`s are now `tabIndex={0}` and activate on Enter/Space, with keydowns originating from the inner dismiss/CTA buttons ignored so one press can't fire both actions. Covered by tests (activation, focusability, inner-button guard, non-activation keys).
- **Suggestion — backstop anchor.** `requestedAt` is captured before `await faucet(address)`, so the 3-minute backstop is anchored to the user's tap rather than the ack.

Housekeeping: the PR is retargeted onto `main` (until #498 merges, the diff here includes that branch's commits — noted in the description), the description now reads "Partially addresses #476", and the attribution footer is removed.

src/main.css
BrianSeong99: how about create a skill.md for this and we can enforce this pattern to agents?

BrianSeong99: Created this PR for the frontend CSS design pattern for agents, maybe once this #515 PR is reviewed and merged to main, @dotsolo could pull the update and update ths implementation in this PR

dotsolo:
Merged latest `origin/main` — resolves the conflicts with #603, which had independently added a funding drawer. How the collision was resolved:

- **This PR's in-card funding lifecycle stays** the faucet card's flow. Main's `FundWalletDrawer` remains in the tree but is no longer opened from the card — same treatment the rebase note already documents for `WalletFundingDrawer`.
- **`faucet()` stays native-MIDEN-only** with the 60s abort timeout (carried from #498, which was closed as superseded by this PR). #603's two-source `FaucetError` aggregation and per-source retry memo were dropped along with the second source.
- **Kept from main:** `faucetFetch` hardening (15s timeout, single 429 Retry-After honor, PoW solve deadline) — now composed with this PR's AbortSignal threading, so `faucetFetch` links a caller signal to its internal timeout controller; the hot-key rotation prompt code + tests (#501); and the pending-notes e2e test ids.
- Locales and CHANGELOG were unioned (both PRs' keys/entries kept).

Verified: `tsc` clean, the 9 touched jest suites pass (153 tests), `eslint` + `lint:i18n` clean.

BrianSeong99 (approved):
LGTM as well

dotsolo:
@WiktorStarczewski gentle ping — re-requested your review after merging latest `main` (twice now: your #603 landed mid-review, then the 1.15.20 release). The resolution summary is in [my earlier comment](https://github.com/0xMiden/wallet/pull/504#issuecomment-5296960712); since your last review the only other changes are the group-claim e2e helper updated for the whole-card tap and the changelog entries moved to the new 1.15.21 (TBD) section. Utkarsh and Brian have approved — yours is the last one outstanding. 🙏

WiktorStarczewski (changes requested):
Re-reviewed at `d06a40e13`. CI is fully green here, i18n is complete across all 14 locales, and the e2e helper correctly follows the CTA→whole-card change. The abort plumbing from round 3 still holds up — the signal is genuinely threaded into both fetches and checked per PoW iteration.

Two things drive this review:

1. **None of the round-3 items were addressed.** Between `277872edc` and here the branch has one non-merge commit (`897fad122`, the e2e helper) plus two merges of `main`. That's consistent with your ping — but the two merge-gating items are still open, including the one I asked you to *decide* rather than necessarily fix.
2. **The `main` merge itself introduced new findings.** Resolving against #603 changed real behavior, and that resolution has not been reviewed before now.

## Round-3 findings — status

| Prior finding | Status |
|---|---|
| 🟠 #1 in-flight guard is a single boolean → funding a 2nd account dead-ends | ❌ open — `HomePrompts.tsx:326` still `if (fundingRef.current) return` |
| 🟠 #2 arrival isn't faucet-specific → any inbound note false-succeeds | ❌ open, and no decision recorded either way |
| 🟡 success branch lacks the account guard the catch has | ❌ open — `HomePrompts.tsx:343-348` |
| 🟡 "Funded!" amount re-mixes baseline notes | ❌ open — `HomePrompts.tsx:470` |
| 🟡 success-beat tests on real timers | ❌ open — `HomePrompts.test.tsx:267,309,352` still 3500/3000ms budgets against a 2400ms beat |
| 🟡 dead code shipped | ❌ worse — see NEW-4 |
| NIT marker-clear writes unguarded | ❌ open — `HomePrompts.tsx:371,389,417` |
| NIT abort comment overclaims | ❌ open — `wallet-prompts.ts:287-291` |
| NIT CHANGELOG describes the dead drawer | ❌ open — moved into 1.15.21, text unchanged |

## New — merge-gating

**NEW-1. 🔴 Faucet failures now surface no reason at all. This un-ships #425/#603, which is on `main`.**

`main` (`HomePrompts.tsx:280` → `FundWalletDrawer.tsx:75`) captures the thrown message and renders it in a `select-text`, `role="alert"` region with **Retry** and **Close**. On this branch the catch block (`HomePrompts.tsx:349-352`) keeps only `setFaucetStatusIndicator('failure')` + a `console.error`; the message is dropped on the floor. The entire user-visible failure surface is a 24×24 red X (`PromptCard.tsx:121-128`) on a card whose title still reads "Fund your wallet". A rate-limited faucet, a 400 "exceeds the maximum claimable amount", and a network outage are now indistinguishable.

The tests that guarded it were deleted rather than ported: `HomePrompts.test.tsx:480` mocks `new Error('rate limited')` and asserts only `data-status === 'failure'`. Nothing asserts any message, because none is produced.

Two aggravating details: the red X never clears (nothing resets the indicator except an account switch or the next tap), and #425 was *specifically* the "show the error and let me retry" issue.

Fix: put the message in the card — `body: errorMessage` on the failure branch of the Faucet override is enough, and it re-uses the existing body slot.

**NEW-2. 🟠 The request-time baseline can be empty or incomplete, so the round-2 baseline gate doesn't actually hold.**

`baselineNoteIds = pendingNoteIds` (`HomePrompts.tsx:332`) is derived from `claimableNotes`, which is SWR data (`Explore.tsx:68`, `useClaimableNotes`) and is `undefined` until the first load resolves. `balances` comes from IndexedDB and lands much earlier, and the card's visibility gate only waits on `balancesLoading` (`HomePrompts.tsx:188`). So the card is tappable while `claimableNotes` is still `undefined` → baseline is `[]`.

Worse, it stays incomplete after the first load: `attachMetadataToNotes` drops notes whose faucet metadata isn't cached yet (`claimable-notes.ts:118-120`, and `:242` on extension), with the explicit comment that they "will appear after metadata is fetched and SWR revalidates".

Sequence: zero-balance account holding one unclaimed note with uncached metadata → tap Fund → baseline `[]` → ack → hero → metadata resolves a moment later → that *pre-existing* note is "beyond the baseline" → "Funds deposited — You have $X ready" and the prompt completes ~30–60s before the mint lands. This is the exact failure the round-2 fix was written to prevent, reached by a different door.

Fix: don't arm a wait against an unloaded set — gate `fundWallet` (and the arrival effect) on `claimableNotes !== undefined`.

This and round-3 #2 are the same defect family. Threading `faucetId` into `PendingNoteValue` and requiring a *new note from the MIDEN faucet* closes both; anything narrower leaves one of them open.

**NEW-3. 🟠 Pre-ack funding state is component-local, and `HomePrompts` unmounts on navigation → a second real mint.**

The marker is only persisted *after* the ack (`HomePrompts.tsx:343`), and `fundingRef` is per component instance. `TabLayout.tsx:217` keys its wrapper `showActionBar ? 'home-group' : pathname`, so any non-home route (Activity, Settings, `/pending-notes`) unmounts `HomeSwipeContainer` and with it `HomePrompts`.

Sequence: tap Fund → `faucet()` in flight (up to 60s; PoW alone is bounded at 30s) → tab to Activity → back to Home → remount restores `idle`/`null`, no marker exists yet, card renders fully tappable → user sees no progress, taps again → second concurrent `faucet()`, second mint.

Fix: hoist the in-flight state out of the component — either persist a `pending` marker *before* `await faucet(...)` (clearing it on rejection), or keep a module-scoped `Map<address, Promise<void>>` in `lib/wallet-prompts.ts` that `fundWallet` joins. The second shape also fixes round-3 #1 for free, and `lib/wallet-funding.ts:22` already has exactly that shape — it's just not wired to anything.

**NEW-4. 🟠 The merge left three dead funding subsystems in the tree, propped up by tests inside the 95% gate.**

- `FundWalletDrawer.tsx` (96 lines) + its 147-line test — main's #603 drawer, now **zero production importers**. Not mentioned in the PR description, which names only `WalletFundingDrawer`.
- `WalletFundingDrawer.tsx` + `lib/wallet-funding.ts` — still *wired* (`TabLayout.tsx:10,20,255`) but `openWalletFunding` has no production caller, so `walletFundingOpen` is permanently `false`. Every `TabLayout` render subscribes to a store that can never change.
- `jest.config.ts:15-22` sets `collectCoverageFrom: ['src/**/*.{ts,tsx}', …]` with a 95% global gate and neither file is in `coveragePathIgnorePatterns` — so ~200 lines of unreachable code are kept green by ~280 lines of tests, including `FundWalletDrawer.test.tsx`'s "renders the error state with the real message, retry, and close wiring", which asserts precisely the behavior NEW-1 removed.
- This PR also *adds* 6 new `walletFunding*` keys × 14 locales that nothing can display, on top of orphaned `fundWallet*` (6), `faucetPromptAction` and `pendingNotesPromptAction`.

Delete both drawers, `lib/wallet-funding`, their tests, the `TabLayout` wiring and the dead keys — or say in the description which one is deliberately parked and why.

## New — non-blocking

- **🟠 Swap's default token pair is now race-dependent.** `tokens.ts:76-84` moves the discovered native asset to the *head* of `getSwapTokens()`, and `SwapManager.tsx:37-38` seeds `useState(() => getSwapTokens()[0]!)` / `[1]!` once at mount. `getNativeAssetIdSync()` is `null` until discovery lands (`native-asset.ts:101-103`), so a cold start into `/swap` defaults to IMIDEN→IETH and a warm one to MIDEN→IMIDEN. Same build, same user, different default pair on a money screen. Pick the defaults by symbol instead of list position, or leave native appended and reorder only for display.
- **🟡 The two new infinite animations have no reduced-motion guard.** `main.css:560-608` adds a 1.4s infinite translate and a 2.2s infinite 360° rotate that run for the entire 30–60s wait. `main.css:329` and `:357` already establish the `@media (prefers-reduced-motion: reduce)` pattern, and the repo's own CLAUDE.md says to keep component-specific animation out of `main.css` and route nontrivial motion through the reduced-motion-aware spring helpers — which is also the substance of the nit @0xnullifier raised on 08-07 and that hasn't been answered.
- **🟡 `faucetHeroActive` can outlive the card it protects.** `HomePrompts.tsx:426` includes `faucetStatusIndicator === 'loading'`, but `showFaucetPrompt` (`:188`) doesn't. If the native auto-consumer settles a note mid-request, `hasBalance` flips, the seeding effect completes the prompt and the faucet card unmounts — yet pending-notes stays suppressed until `faucet()` settles, leaving an empty carousel for up to 60s. Gate on `showFaucetPrompt && (…)`.
- **🟡 The backstop isn't clamped to its own constant.** `HomePrompts.tsx:414` computes `requestedAt + 3min − now` from a persisted wall-clock stamp. A backward clock step (NTP correction, manual change) makes `remainingMs` arbitrarily large and wedges the hero — and pending-notes suppression — well past 3 minutes. `Math.min(FAUCET_FUNDS_ARRIVAL_TIMEOUT_MS, …)`, and reject a future `requestedAt` when reading the marker.
- **🟡 CHANGELOG.** The 1.15.21 CHANGE bullet (`CHANGELOG.md:11`) still describes the drawer — "opening the funding drawer starts the native faucet request automatically and keeps its loading, success, or **detailed error state** correct" — which is both unreachable code and, per NEW-1, a feature the shipped flow no longer has. It also contradicts the FIX bullet four lines above it in the same section. Separately, 1.15.20's released bullet "Retry now re-mints only the source that actually failed" is no longer true of the code; worth a line in 1.15.21 noting the two-source funding was removed (the IMIDEN half is already covered). Section placement is correct.
- **🟡 A slow mint can now be minted twice.** `main` completed the prompt on ack, so the card was gone. Here the backstop (`HomePrompts.tsx:411-420`) clears the wait and the marker at 3 minutes without completing the prompt, returning a fully actionable card. A mint that merely took >3 minutes gets a second one on the retry the UI is inviting. Low harm (devnet test tokens) and it's the price of having any in-session recovery at all — but worth an explicit decision, e.g. keeping a "already requested" marker past the backstop.
- **NIT — the abort-forwarding test asserts nothing.** `faucet-api.test.ts:238` expects `{ signal: controller.signal }`, but `faucetFetch` deliberately passes its *internal* controller's signal (`faucet-api.ts:67`). It passes only because Jest structurally equates two non-aborted `AbortSignal`s — so it would keep passing if the linking were replaced by a naive `init.signal` pass-through, which would break the 15s timeout. Assert behavior (abort → rejection) like the `getPowChallenge` sibling test at `:84`.
- **NIT — the 429 back-off sleep ignores the caller's abort.** `faucet-api.ts:79` sleeps up to `MAX_RETRY_AFTER_MS` (30s) in a bare `setTimeout`, outside the linking in `attempt()`. `Promise.race` still rejects the caller on time, so nothing hangs — but the "aborts the underlying work" claim is delayed by up to 30s.
- **NIT — clearing the marker writes `null` rather than removing the key**, leaving one dead `faucet_funding_v2:<address>` entry per funded account.
- **NIT — `role="button"` on a container holding real `<button>`s.** ARIA makes a button's children presentational, so the dismiss X isn't reliably exposed. And when the hero engages, `onClick` becomes `undefined` → `tabIndex` disappears → a keyboard user who just activated the card loses focus to `<body>`. Consider `aria-disabled` + a retained `tabIndex` during the hero.

## Verdict

Still requesting changes, on: **NEW-1** (error surface — it's a straight regression of shipped `main` behavior), **NEW-2 + round-3 #2** (arrival correctness; one fix closes both), **NEW-3 + round-3 #1** (in-flight state hoisted out of the component), and **NEW-4** (delete the dead subsystems or document them). Everything else can ride along or follow up.

Also worth noting: nothing here changed since 08-11 in the funding code itself, so if you'd rather split this — land the lifecycle with the error surface restored and take the arrival-specificity work as a follow-up issue — say so and I'll review it that way.

dotsolo:
@WiktorStarczewski all four gating findings are addressed in `bdca8565` (net −1,269 lines), full-fix route — no split needed. Finding-by-finding:

**NEW-1 (error surface):** the catch now stores the thrown message (per-account, like the indicator) and the Faucet override renders it through the card's existing `body` slot on the failure branch, so a rate limit, a rejected amount, and an outage read differently. Cleared on retry and on account switch. The failure test now asserts the actual message renders (`rate limited`), not just `data-status`.

**NEW-3 + round-3 #1 (in-flight state):** `faucet()` now joins per-address at module scope (`Map<address, Promise<void>>` in `lib/wallet-prompts.ts`, exposed via `getInFlightFaucetRequest`), so a tap from a remount joins the running request instead of minting twice, and funding one account never blocks another. The marker is persisted **before** the request (cleared on rejection), so a returning user sees the Funding hero, not an idle tappable card; a remounted card also re-attaches to the in-flight promise to paint loading and any failure. `fundingRef` is gone. Tests: concurrent join, per-address independence, post-settle re-fund, remount re-attach + failure paint, cross-account funding.

**NEW-2 + round-3 #2 (arrival correctness):** `PendingNoteValue` now carries `faucetId` (the runtime objects always had it). `fundWallet` and the arrival effect both gate on `claimableNotes !== undefined`, and arrival requires a beyond-baseline note **from the native MIDEN faucet** (`useMidenFaucetId`); the balance path is unchanged. Your metadata-resolves-late sequence and round-3's any-inbound-note case are both covered by tests (undefined-tap no-op; non-native new note holds the hero, native note completes).

**NEW-4 (dead subsystems):** `FundWalletDrawer`, `WalletFundingDrawer`, `lib/wallet-funding`, all three test files, and the TabLayout wiring are deleted; the 14 orphaned keys (`walletFunding*`, `fundWallet*`, `faucetPromptAction`, `pendingNotesPromptAction`) are removed from all 15 locale files. The 1.15.21 CHANGE bullet is rewritten — no drawer wording, and it now notes the two-source retry bookkeeping from 1.15.20 is retired with the second source.

Ride-alongs from your round-3 list: the success branch got the same account guard as the catch, the marker clears are `.catch`-logged, and the abort comment no longer overclaims (it names the 429-sleep lag).

Everything else from the non-blocking list is tracked in #709 so it doesn't ride this diff.

Verified: `tsc` 0 errors, full jest 7489/7489 across 537 suites, coverage 95.35% functions / 96.34% lines (gate ≥95%), `eslint` + `lint:i18n` + prettier clean.

PR #507, opened 2026-08-06
fix: keep transaction receipt open + proper consume receipt

0xnullifier:
## What

- **No more auto-navigate home after a transaction completes.** The Generating Transaction screen used to redirect to Home ~10s after the tx reached a terminal state; users like to stay and read the receipt, so the success/failure view now stays up until the user taps Done (or Hide). Removed the now-dead auto-close setting plumbing (`isAutoCloseEnabled`/`setAutoCloseSetting`, storage key/default, orphaned `AutoCloseToggle` selector — nothing in the settings UI exposed it anymore).
- **Consume (claim) transactions get their own summary + receipt.**
  - In-progress pill: `{amount} {symbol} → Consumed` (renders nothing when the consume carries no amount).
  - Success receipt: address row relabeled **From** (the note sender), **Total Consumed** instead of "Total Paid", and a new **Notes Consumed** row listing the claimed note ids (batch-aware via `noteIds`).
- **Receipt hash row renamed "Transaction ID"** (was "Source TX") for all transaction types.
- Reworded the sr-only "do not close this window" copy that promised auto-close/auto-navigation.

## Testing

- Updated/added unit tests: no auto-navigation after terminal state, consume badge variant, consume receipt labels + notes row (97 tests in the generating-transaction suites pass).
- `yarn lint`, `yarn lint:i18n`, `yarn format`, `tsc --noEmit` all clean.


WiktorStarczewski (commented):
Minor (i18n): the new `notesConsumed` key leaves the leading noun **"Notes" in English** across the 12 non-English locales (de "Notes Verbraucht", ru "Notes Израсходовано", ja "Notes 消費", …), while its two sibling keys added in the same hunk (`totalConsumed`, `consumed`) are fully localized (e.g. de "Gesamtverbrauch" / "Verbraucht"). So the receipt's "Notes Consumed" row reads half-translated in those languages.

If "Notes" is a deliberate Miden domain term, keep it English consistently across the locales; otherwise translate the whole phrase per locale.

PR #509, opened 2026-08-06
fix(send): validate recipient addresses with a strict bech32 decode

0xnullifier:
## Problem

It was possible to enter an invalid bech32-encoded account ID in the Send screen. Miden-address validation was only a prefix check (`startsWith('mtst1'|'mm1'|'mdev1')`), so a typo'd address sailed through the UI and only failed deep in the transaction pipeline — after the "Guardian approved" stage — surfacing as a bech32 error in the Activity overview. The prefix check also accepted all three network prefixes at once, so a mainnet address passed validation on a testnet build.

## Fix

- `isValidMidenAddress` (`src/utils/miden.ts`) now runs the SDK's full bech32 decode (`Address.fromBech32`, the WASM equivalent of the Rust `Address::decode`). It returns `true` for a valid address of the current network and throws `MidenAddressError` with `reason: 'invalid' | 'wrong-network'` otherwise.
- The Send screen maps the two failure reasons to distinct messages: "Invalid Miden account ID" for a typo, and a new "This address belongs to a different Miden network" for a well-formed address of another network (mm/mtst/mdev prefix vs. the build's `MIDEN_NETWORK`; localnet encodes with the testnet prefix, mirroring `getNetworkId()`).
- Validation now runs on all three recipient entry paths: typing, QR scan, and contact selection (previously scan/contact only had the self-send guard).
- The QR-side validator (`src/lib/qr/format.ts`) also does a real decode instead of prefix + length; a wrong-network scan is let through to the field so the send screen can show the specific message.

Decode success is a sufficient check: verified against SDK 0.15.9 that both address forms — with the `_…` routing-parameters suffix (BasicWallet interface) and the bare account-id form — decode to the same note tag, and the routing suffix is fully reconstructible from the account id + interface, so no separate routing/note-tag validation is needed.

## Tests

- Rewrote `src/utils/miden.test.ts` for the strict validator (valid / wrong-network / decode-failure paths).
- `__mocks__/wasmMock.js` gained a loose `Address.fromBech32` stand-in (HRP + bech32 charset + length) that individual tests can override.
- Fixed the QR format test's mainnet fixture (`m1…` → the real `mm1…` prefix).
- 281 tests green across send-flow, qr, utils, and AddressBook suites; `tsc`, `yarn lint`, `yarn lint:i18n` clean.


WiktorStarczewski (approved):
LGTM

PR #510, opened 2026-08-06
fix(private-send): adopt web-sdk#263 sendPrivateOutput (expected-height hint)

WiktorStarczewski:
## Summary

Adopt the SDK's `sendPrivateOutput` (web-sdk#263) for private-note delivery. The SDK now derives the recipient's transport block hint from the note's **expected height** (the chain tip when its transaction was submitted), so delivery is correct regardless of this client's sync height or relay timing.

This is the proper SDK-level fix behind the wallet workaround merged in #502 (which relayed before the commit wait to keep a sync-height hint below the commitment). With the hint derived from expected height, timing no longer matters; the wallet just calls the new convenience.

## Changes

- `miden-client-interface.ts`: `notes.sendPrivate({ note, to })` → `notes.sendPrivateOutput({ noteId, to })`.
- `complete.ts`: refreshed the two relay-site comments — the sync-height rationale no longer applies (the relay-before-commit ordering from #502 is retained but is now redundant for correctness).
- `CHANGELOG.md`: 1.15.20 entry updated to describe the expected-height fix.

## Verification

The SDK side is covered by web-sdk#263's cross-client regression test (relays via `sendPrivateOutput` after syncing past the note's commitment; fails if the hint regresses to sync height). This PR's CI builds the wallet against #263 via the marker below and runs the local blockchain e2e (including the 500ms fast-block `send-private` leg added in #502).

Web SDK PR: #263

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- [FIX][all] **Private-note delivery no longer silently fails on fast chains.** The wallet relayed a private note's transport hint *after* waiting for the transaction to commit, so the hint (the client's sync height) could already be at or past the note's commitment block — the recipient scans forward from the hint and would never find the note. The note is now relayed *before* the commit wait (the SDK's prompt-relay flow), so the hint stays below the commitment block and the recipient picks the note up when it lands; the commit wait is preserved so `Completed` status still requires on-chain confirmation. Latent on testnet (slow blocks kept the sync height ≈ the commitment block at relay time); reproduced deterministically on a fast devnet.
+- [FIX][all] **Private-note delivery no longer silently fails on fast chains.** A private note's transport "block hint" — the block the recipient scans forward from for the note's on-chain commitment — was the client's current sync height, which could already sit past the commitment once the wallet had synced past the note, so the recipient scanned past it and never found the note. The wallet now relays via the SDK's `sendPrivateOutput`, which derives the hint from the note's expected height (the chain tip when its transaction was submitted), so delivery is correct regardless of sync height. Latent on testnet (slow blocks kept the sync height ≈ the commitment block at relay time); reproduced deterministically on a fast devnet. Requires web-sdk#263.
 
 ## 1.15.19 (2026-08-05)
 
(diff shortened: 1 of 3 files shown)


WiktorStarczewski:
Closing: superseded on `next`, and not being taken to `main` for now.

This PR's whole payload — relaying through `notes.sendPrivateOutput` so the
recipient's scan-start block comes from the note's stored `expected_height`
instead of the sender's live sync height — is already present on `next`:

- `src/lib/miden/sdk/miden-client-interface.ts:522` calls
  `this.client.notes.sendPrivateOutput({ noteId: note.id().toString(), to })`.
- The only remaining occurrence of `notes.sendPrivate(` anywhere on `next` is a
  comment in `miden-client-proxy.ts` recording what 0.15 used to do. There is no
  live call site left.
- The behaviour is locked in by a test (`miden-client-interface.test.ts:342`
  asserts `sendPrivateOutput` is called with the note id and recipient).
- The `complete.ts` rationale this PR rewrote has already been updated on `next`
  to the correct explanation: the relay-before-commit ORDER was load-bearing under
  0.15 because the hint moved with sync, and under 0.16 the hint comes from
  `expected_height`, which does not.

`next` got this natively as part of the 0.16 SDK migration rather than as a
cherry-pick: 0.16's `notes.sendPrivate` requires an explicit `scanAfterBlockNum`,
so adopting `sendPrivateOutput` was forced by the migration. Rebasing this branch
onto `next` would therefore produce an empty diff.

It is not being forward-ported to `main` either. `main` still relays via
`notes.sendPrivate({ note, to })`, but the 0.15.9 SDK pinned there does not expose
`sendPrivateOutput` at all, so this change cannot land on `main` without the
web-sdk bump — which is what the `next` line already is.

Worth recording for anyone reading this later, since the original description was
imprecise about it: 0.15.9 did NOT relay without a hint. It derived one from the
sender's live `get_sync_height()` and called `send_private_note_with_block_hint`.
The bug was that the hint could OVERSHOOT the note's commitment once the sender
synced past it, and a hint above the commitment is never scanned back to — which
is the silent non-delivery this PR set out to fix. The diagnosis holds; only the
"no hint at all" framing was wrong.

Follow-up work on private-note delivery continues in #765.

PR #511, opened 2026-08-06
Fund-wallet flow: full funding/success lifecycle + prompt card redesign

dotsolo:
> [!NOTE]
> Re-opened continuation of #504 — the head branch was renamed `fundwalletloadfix` → `fix/fund-wallet-loading` and rebased onto `fix/remove-imiden-swap-faucet`, which closed the original PR. Content is unchanged apart from the rebase.

## Problem

On iOS, tapping **Fund this wallet** acked the faucet request within seconds, but the minted note only becomes visible after chain inclusion + a client sync (~30–60s). The prompt card flashed success on the ack and vanished, so the Home screen looked completely idle for the whole gap until the note notification appeared.

## What changed

**Funding lifecycle (HomePrompts):**
- The faucet card itself is now the tap target (no separate CTA button).
- Tapping enters a **"Funding" hero**: animated hourglass tile, indeterminate progress bar along the card's bottom edge, and *"Your funds will be available shortly"*.
- When the funds land (claimable note or balance), a green **"Funds deposited — You have $X ready"** success beat plays (~2.4s), then the prompt completes and hands off to the pending-notes card.
- The pending-notes card **holds back while the hero is on stage** — it sorts first in the carousel and previously would have pushed the success state off-screen the instant the note arrived.
- The wait **survives remounts and app restarts** via a persisted request marker (`faucet_funding_requested_at_v1`), and falls back to the actionable card after 3 minutes if the mint never lands.
- The pending-notes card's "See Now" button was folded into a whole-card tap as well.

**PromptCard:**
- New optional icon slot (coins icon on the faucet card), filled pill CTA, corner dismiss X that hides while a request is in flight, press-in feedback (`active:scale-[0.98]`) on tappable cards, and a reusable `hero` takeover mode (icon tile + wordmark + sub-line lockup).
- New `hourglass` icon added to the v2 icon set.

**Polish:**
- Home "Assets" heading now uses the heading font (Nunito) — it was falling back to Inter.
- Onboarding confirmation screen: bold Nunito heading, Nunito body copy, and the daily-reminder note styled as a pill (also fixes a stray `font-medium` JSX attribute that was never applied).

## Rebase note

This branch is now stacked on `fix/remove-imiden-swap-faucet`. Where the two collided (`HomePrompts`, its tests, locales), this branch's in-card funding lifecycle supersedes the base's `WalletFundingDrawer` wiring — the drawer and `lib/wallet-funding` remain in the tree but are no longer opened from the faucet card. The base's native-MIDEN-only `faucet()` and swap-token ordering are preserved and compose with this flow.

## Testing

- 95 tests pass across both branches' funding-flow suites (`HomePrompts`, `wallet-prompts`, `wallet-funding`, `WalletFundingDrawer`, `TabLayout`), including new coverage for: card-tap funding, the Funding→Funded! sequence, arrival via notes and via balance, the persisted-marker resume path, pending-notes suppression during the success beat, and tap-to-retry after failure.
- Verified end-to-end on the iPhone 17 simulator against testnet: onboard → tap card → Funding hero through the ~30s gap → Funds deposited beat → pending-notes hand-off.


dotsolo:
Superseded — the original #504 was restored and reopened with the same rebased content (base: `fix/remove-imiden-swap-faucet`).

PR #513, opened 2026-08-07
fix: cache token metadata fetched during sync

0xnullifier:
## Summary

- persist successful token metadata RPC results to the existing tokens_base_metadata cache
- serialize cache writes so parallel faucet lookups preserve every metadata entry
- keep successful metadata usable when storage persistence is unavailable

## User impact

Background syncs now reuse previously fetched faucet metadata instead of issuing repeated GetAccount requests for the same token.

## Testing

- yarn test src/lib/miden/metadata/fetch.test.ts --runInBand
- yarn ts


WiktorStarczewski (commented):
Minor — a transient miss can poison the cache (not blocking):

`src/lib/miden/metadata/fetch.ts` — the transient public-account branch (`!underlyingAccount && account.isPublic()`, commented *"should not happen … but in case it does"*) now flows through `cacheTokenMetadata`, persisting `DEFAULT_TOKEN_METADATA` into `tokens_base_metadata`. The top-of-function cache check then treats it as a hit and never retries, so a token whose node lookup transiently returned an empty account shows "Unknown" permanently.

Caching is correct for the deterministic *private* sub-case but wrong for the transient *public* one — consider returning `DEFAULT` **without** persisting when `account.isPublic()` is true. (This was already partially reachable pre-PR via `useAssetMetadata` persisting to the same key; the change just extends it to the sync-manager path and `fetch.ts` itself.)

WiktorStarczewski (changes requested):
Thanks — this is a follow-up from a deeper multi-pass review of the metadata-persistence change. The direction is good, but I'd like one thing addressed before merge.

### Requesting changes

**The concurrent-write serialization — the PR's stated centerpiece — isn't tested, and the new queue doesn't actually serialize against the pre-existing writer.**

1. **Untested (important).** The only new test (`fetch.test.ts:122-147`) awaits two *sequential* fetches of the *same* `assetId`; the second is an early cache-return hit (your own `mockGetAccountDetails` called-once assertion proves it never re-enters the persist path). So exactly one persist write happens and `metadataStorageWriteQueue`'s `.then()`-chaining is never exercised. A regression that weakens the chaining (e.g. reading the queue after an `await`, or dropping the `.then()`) would ship green. Please add a test that fires `fetchTokenMetadata('asset-a')` and `fetchTokenMetadata('asset-b')` concurrently via `Promise.all` with distinct mocked account details, and asserts the final stored object contains **both** entries (no lost update).

2. **Second writer still races (minor, but it undercuts the fix).** `persistTokenMetadata` adds a *second* independent read-modify-write path to `tokens_base_metadata`, serialized only by its own `metadataStorageWriteQueue`. The pre-existing `setTokensBaseMetadata` (`assets.tsx`, serialized by its own `setAllTokensBaseMetadataQueue`) writes the same key, and `putToStorage` does a full `set({[key]: value})`. The two queues share no lock, so a background-sync write via your path and a foreground `useAssetMetadata`/`fetchBalances` write via the other can still clobber each other — the exact lost-update the PR sets out to prevent. Consider routing all `tokens_base_metadata` writes through one shared serialization point (e.g. have `persistTokenMetadata` reuse `setTokensBaseMetadata`).

### Minor / non-blocking

- The three `DEFAULT_TOKEN_METADATA` fallback branches now persist via `cacheTokenMetadata`, but no failure-path test asserts on `mockPutToStorage`. In particular, permanently caching `DEFAULT` for the *transient public* "should not happen" branch (`isPublic() && account()===null` during an RPC lag) would pin the token to "Unknown" forever. Worth an explicit assertion — and returning `DEFAULT` *without* persisting for that transient case.
- The `cacheTokenMetadata` try/catch resilience path carries `/* c8 ignore */` and is never exercised (no test rejects `putToStorage`). A test that rejects the write and asserts `fetchTokenMetadata` still resolves with the RPC result would lock it in.

WiktorStarczewski (approved):
LGTM — thanks for the thorough follow-up. The concurrent-write serialization is now covered by a real lost-update test (`asset-a`/`asset-b` via `Promise.all`, asserting both survive), the transient-public branch returns default without poisoning the cache (with a retry test), and both writers now share a single `updateTokensBaseMetadata` queue so the within-realm races are gone. The remaining SW↔page cross-realm write window is the deeper architectural item I'd flagged as non-blocking — fine as a follow-up.

PR #518, opened 2026-08-09
fix(dapp-browser): scope injected postMessage to the page origin

maymuneth:
## What

`sendToNative` in the injected dApp-browser script falls back to `window.postMessage(..., '*')` when the Capacitor `mobileApp` bridge isn't present.

## User impact

This script is injected into arbitrary third-party dApp pages. A wildcard target origin means every message listener on that page — including cross-origin iframes the dApp embeds — receives the wallet request type and payload. The comment directly above (lines 69–71) already states the payload must not be exposed on the dApp page, so the wildcard broadcast works against that intent.

## Change

Target `window.location.origin` instead of `'*'`. The intended listener lives on the same page, so nothing functional changes.

This aligns the file with the patterns already used elsewhere in the repo:

- `src/lib/adapter/client.ts:309` posts to `window.location.origin`
- `src/contentScript.ts:106-108` rejects a missing or `'*'` target origin

## Testing

I wasn't able to run the suite locally, so this is reasoned from the source rather than executed:

- The changed line is the non-Capacitor fallback, reached only when `window.mobileApp` is absent, so native builds don't touch it
- The intended listener runs on the same page, and `window.location.origin` matches the page's own origin, so delivery is unchanged
- The two other places in this repo that post from injected or content-script context already scope the origin: `src/lib/adapter/client.ts:309` posts to `window.location.origin`, and `src/contentScript.ts:106-108` rejects a target origin that is missing or `'*'`

Happy to add a test asserting the target origin if that's useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 
+- [FIX][mobile] **The injected dApp bridge no longer broadcasts wallet request payloads to every listener on the page.** `sendToNative`'s non-Capacitor fallback posted the request with `targetOrigin: '*'`. That script is injected into arbitrary third-party dApp pages, so any listener there — including cross-origin iframes — could read the request type and payload. The target origin is now scoped to `window.location.origin`, matching the existing pattern in `lib/adapter/client.ts` and the guard in `contentScript.ts`. The intended listener is on the same page, so behaviour is unchanged.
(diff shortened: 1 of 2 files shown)


canoplos:
scoping the fallback `postMessage` to `window.location.origin` rather than `'*'` avoids leaking the request payload to other listeners (incl. cross-origin iframes) on the injected dApp page, and it matches the existing `lib/adapter/client.ts` pattern. 

Minor nit: the new comment block / `postMessage` line indentation is inconsistent with the surrounding code and will probably trip Prettier/ESLint — worth a quick reformat. Otherwise LGTM.

maymuneth:
Fixed the indentation, the block now lines up with the surrounding code. I also noticed my earlier conflict resolution had dropped a batch of CHANGELOG entries from main, so I restored the file and re-added only my own line. Thanks for the review.

PR #534, opened 2026-08-10
Move history-detail polling to app-root watchers

0xnullifier:
> **Stacked on #⁠— `feat/bridge-via-deposit-address`** — only the last commit (`a206efdb`) belongs to this PR; the base branch carries the in-progress deposit-address bridge work it builds on.

## What

Moves every polling loop out of the activity detail page (`HistoryDetails`) into always-mounted app-root watchers, and turns the page into a passive Dexie `liveQuery` observer.

### App root (new)

- **`EarnIntentWatcher`** (`src/lib/epoch/EarnIntentWatcher.tsx`, mounted in `provider.tsx`): every 15s, a cheap Dexie scan for non-terminal `earn-deposit` / `earn-withdraw` rows; when a row has no live poll, it dynamic-imports and re-runs the existing `reconcileEarnDeposits` / `reconcileEarnWithdrawals`. Replaces both the once-per-session reconcile on Home mount (removed from `Explore.tsx`) and the detail page's poller kicks. Steady state is one IndexedDB scan per tick — zero network, zero heavy imports.
- **`poll-registry`** (`src/lib/epoch/poll-registry.ts`): `pollEarnIntentStatus` / `pollEarnWithdrawDelivery` now self-dedupe by nonce (claim on start, release on every stop), so any context can safely re-kick them. The page's `withdrawPollNonceRef` / `depositPollNonceRef` hacks are gone.
- **`SwapOrderTrackingManager`** + **`order-tracking-store`**: the `trackOrderId` lineage poll — with its exact exponential backoff / give-up-after-20 semantics, one WASM-lock acquisition at a time — runs at the root and publishes into an in-memory zustand store keyed by orderId. Opening the detail page calls `requestRefresh`, preserving the old "reopening restarts a given-up poll" behavior.

### Detail page

- Observes its row via `useTransactionRow` (liveQuery). The pending-status reload loop, both earn reload loops, the swap lineage poll, and the bounded settlement-notes poll are deleted.
- Settlement notes become a push-based `useSwapSettlementNotes` liveQuery hook (no more 20-poll cap for manual-claim orders).
- Cancel/retry no longer re-fetch — the Dexie write pushes the updated row. `BridgeClaimSection.onUpdated` became optional for the same reason (its internal 8s pollers are out of scope and unchanged).

## Why

All of the page's pollers died with the page: an earn intent could sit stranded on "pending" until the user happened to revisit Home or the detail page, and swap tracking restarted from scratch on every visit. Root-mounted watchers keep status advancing on any screen, and liveQuery makes the page reflect writes instantly instead of on 2–3s refetch intervals.

## Behavior notes

- Swap-order give-up is now per-session-per-order instead of per-page-visit (revived on page open).
- On the extension, an unresolved swap order can now init the popup's WASM client without the detail page being open (gated on such an order actually existing in Dexie).
- Watchers run only while the popup/app is open — the same lifetime the page-mounted pollers had, so no extension regression.

## Testing

- `HistoryDetails.test.tsx` reworked to drive the row / store / hook instead of timers (44 tests).
- New suites: `poll-registry.test.ts`, `EarnIntentWatcher.test.tsx`, `SwapOrderTrackingManager.test.tsx` (the backoff/give-up timing tests were ported here).
- Epoch poller suites gained dedupe/key-release cases.
- `tsc --noEmit`, `eslint --max-warnings 0` on touched files, and 502 tests across `src/lib/epoch`, `src/lib/miden/front`, `src/app/templates/history` all pass.


0xnullifier:
Replaced by #812, which is based on next.

PR #515, opened 2026-08-07
docs: add shared wallet frontend skill

BrianSeong99:
## Summary

Adds one vendor-neutral Miden Wallet frontend skill and exposes it to both Claude and Codex through project-local links. The skill codifies component reuse, semantic styling, motion, accessibility, haptics, platform constraints, and verification.

The repository instructions now carry the same concise frontend baseline, while the detailed guidance remains in the shared skill.

## Validation

- `quick_validate.py` passes for `skills/miden-wallet-frontend`.
- Both `.agents/skills/miden-wallet-frontend` and `.claude/skills/miden-wallet-frontend` resolve to the canonical skill directory.
- `git diff --check` and post-commit `git show --check` pass.
- Checked the PR #504 animation case, component selection, global CSS boundary, reduced-motion, haptic, and mobile-verification guidance.

## Test plan

- [x] Validate skill frontmatter and structure
- [x] Verify shared Claude and Codex skill links
- [x] Verify no vendor-specific metadata is present in the canonical skill
- [x] Review staged and committed diffs for whitespace errors

## Out of scope

- Runtime UI changes or frontend-debt cleanup
- Dependency or lockfile changes
- Changes or comments on PR #504

PR #542, opened 2026-08-10
docs: prefer shared wallet buttons in frontend skill

BrianSeong99:
## Summary

Refines the shared frontend skill after review feedback on #515. Standard wallet actions now default to `components/Button`; bare native buttons are reserved for compact, icon-only, or inline controls that the full CTA cannot represent.

## Validation

- `quick_validate.py` passes for `skills/miden-wallet-frontend`.
- `git diff --check` and post-commit `git show --check` pass.

## Out of scope

- Runtime UI component changes
- Replacing existing controls across the application
- Dependency or lockfile changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


0xnullifier (approved):
LGTM

PR #545, opened 2026-08-11
fix(home): make Settings reachable from the home page

### LINKED ISSUE #434: Settings not accessible from the home page (only via Explore/Activity tabs)
## Symptom

Settings is not reachable from the wallet home page. The settings gear only appears on the Explore and Activity tabs, so from home the user has to navigate to another tab first to open settings.

## Expected

Settings should be accessible from the home page like on the other tabs.

## Notes

The gear button is part of `TabHeader` (`src/components/ui/TabHeader.tsx`), which is only rendered by `AllHistory.tsx` (Activity) and the dApp launcher (Explore); the home screen's header has no settings entry.

Feedback DB ref: ticket 128 (iOS, but applies to all platforms).

WiktorStarczewski:
Closes #434

## Problem

Settings was not reachable from the wallet **home** page. The settings gear only appears on the Explore and Activity tabs (both carry it via `TabHeader`), so from Home the user had to switch to another tab first just to open Settings.

## Fix

Added the same settings gear button to the home/overview page (`src/app/pages/Explore.tsx`), as a fixed row pinned above the scrolling asset list (outside the scroll region, so it stays put). It reuses `TabHeader`'s exact button — same `settings` aria-label, `bg-gray-25` circular styling, and `hapticLight()` + `navigate('/settings')` behavior — so it's visually and behaviorally identical to the gear on the other tabs.

Scoped to the overview pane only (not the Send/Receive/Earn/Swap panes of the home carousel).

## Test

`Explore.test.tsx` gains a test asserting the home page renders a `settings` button that navigates to `/settings` on tap (it fails without the fix — "Unable to find a label with the text of: settings"). Added `isHapticFeedbackEnabled: () => false` to the existing `lib/settings/helpers` mock so the shared `hapticLight()` call is a clean no-op under test.

## Notes

`AccountsDrawer` (opened from the balance card's "more" affordance) already linked to `/settings`, so the route/pattern is well established — this just surfaces the direct one-tap gear the issue asked for.


WiktorStarczewski:
Closing per design review. Visual verification in Chrome (extension popup width) showed this adds a **second gear** to the home screen: the BalanceCard already has an "account options" gear (bottom-right of the card) that opens the accounts drawer — which itself contains a **Settings** link. A standalone top-right settings gear sits right next to it, so the home screen ends up with two near-identical gear icons doing different things.

Since Settings is already reachable from Home (account-options gear → drawer → Settings), #434 is effectively addressed without new UI. Not merging a redundant/confusing second gear.

PR #550, opened 2026-08-11
fix(tx): keep the completion screen up instead of auto-redirecting home

### LINKED ISSUE #472: Payment Sent screen redirects home before the user chooses an action
From feedback ticket #133 (https://app.notion.com/3ae99411cf9081f2860bfb997b241f1e). Reported on iOS.

## Problem

After a transaction, the **Payment Sent!** screen shows transaction information plus **Done** and **View in activities**, but automatically redirects to Home after a few seconds without user input.

This removes the information before the user can review it or choose the Activity destination.

## Expected

Keep the completion screen visible until the user selects an action, or provide a clearly communicated countdown that preserves an accessible route to the completed transaction.

## Related

#170 addressed the opposite historical behavior in the extension. This report concerns an iOS auto-redirect that happens before the user acts.

Feedback DB ref: ticket 133.

WiktorStarczewski:
Closes #472

## Problem

After a send/swap completed, the transaction screen redirected **Home** on its own ~10 seconds later — removing the receipt and the **Done** / **View activities** choice before the user could review or act. (Reported on iOS.)

## Root cause

`GeneratingTransaction.tsx` schedules an auto-close `AUTO_CLOSE_DELAY_MS` (10s) after completion, gated on `isAutoCloseEnabled()`. That setting defaults to `DEFAULT_AUTO_CLOSE`, which was `true` — and there is **no UI to turn it off** (`setAutoCloseSetting` exists but is never called anywhere). So every completion auto-dismissed, on every platform. It's the only auto-redirect path (`isAutoCloseEnabled()` has a single consumer; the success screens' `navigate('/history')` are the user-tapped "View activities" action).

## Fix

Default `DEFAULT_AUTO_CLOSE` to `false` (`src/lib/settings/constants.ts`) — the completion screen now stays until the user picks an action, matching the issue's primary expected behavior. The setting and `setAutoCloseSetting` remain, so auto-dismiss can still be opted into (e.g. via a future toggle); the change is just the default.

I chose "keep the screen visible until the user acts" over the issue's alternative (a visible countdown) as the simpler, lower-risk option — happy to switch to a countdown if preferred.

## Tests

- `helpers.test.ts`: `isAutoCloseEnabled()` now defaults to `false` (fails before the change).
- `GeneratingTransaction.test.tsx`: the screen does **not** redirect Home after 30s when auto-close is disabled (the new default); the existing "auto-closes when enabled" test still passes, so the screen still honors the setting.

## Notes / follow-ups (out of scope)

- **Background sync stays suppressed while on the completion screen.** `useSyncTrigger` suppresses background sync while the URL is on `generating-transaction`; the 10s auto-close previously bounded that window. Sync resumes as soon as the user navigates away (Done / View activities), and the receipt shows static result data — but a user who lingers won't pick up new notes/balance until they leave. Inherent to "keep the screen up"; a follow-up could resume sync once the tx is complete.
- **A screen-reader, processing-only line** (`doNotCloseWindowNavigateHome` / `doNotCloseWindowAutoClose`) still says the window will auto-close / return home "after the transaction is generated". With auto-dismiss off that's no longer accurate; updating the copy (and collapsing the now-vestigial `keepOpen` distinction) is a small follow-up that touches the i18n pipeline.
- The **"Auto Close" toggle placeholder** in Advanced Settings (`onChange` commented out, hardcoded `checked`) is pre-existing dead UI; wiring it to `isAutoCloseEnabled`/`setAutoCloseSetting` would give users the opt-in.


0xnullifier:
duplicate of #507

PR #552, opened 2026-08-11
fix(i18n): use Cerrar for Spanish Close action

### LINKED ISSUE #469: Fix incomplete and incorrect Spanish localization
From feedback ticket #127 (https://app.notion.com/3ae99411cf9081b9bc4ccd5d01051150). Reported on iOS.

## Problem

With the phone configured in Spanish, Bread displays a mix of English and Spanish. Some Spanish strings are not capitalized consistently, and at least one translation is incorrect: **Close** is rendered as **Acerca**, which means “about/near,” not the action to close.

## Expected

The Spanish locale should not fall back to English for shipped strings, should follow consistent capitalization, and should use context-appropriate translations for actions.

## Validation

Audit the full iOS wallet flow in Spanish and add coverage for missing keys and ambiguous source strings such as “Close.”

Feedback DB ref: ticket 127.

doresa0:
## Summary

Spanish `close` was `Cerca` (near). Use `Cerrar`, and capitalize `Resincronizar`.

## Testing

- Checked `public/_locales/es/es.json` and `messages.json`

Closes #469


0xnullifier:
yeah I think single out changes to es.json is not required it would be better if we could come up with a better generalised solution

